PR #3, opened 2016-05-12
use dirname(@__FILE__) rather than Pkg.dir

tkelman:
allows installing the package elsewhere

While I'm opening this and trying to familiarize myself with this code, the per-benchmark time tolerance probably doesn't need to be reported to `.00%` level of precision in the Nanosoldier reports. Where does that specific printing live?

### DIFF
--- a/src/BenchmarkTools.jl
+++ b/src/BenchmarkTools.jl
@@ -69,6 +69,6 @@ export tune!,
 # Plotting Facilities (loaded on demand) #
 ##########################################
 
-loadplotting() = include(joinpath(Pkg.dir("BenchmarkTools"), "src", "plotting.jl"))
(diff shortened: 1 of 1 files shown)


jrevels:
Oh nice, I didn't know about `@__FILE__`. That'll come in handy in the future.

> Where does that specific printing live?

Most of the Nanosoldier [report printing is in Nanosoldier.jl](https://github.com/JuliaCI/Nanosoldier.jl/blob/master/src/jobs/BenchmarkJob.jl#L462), but it calls [this BenchmarkTools method for pretty printing percentages](https://github.com/JuliaCI/BenchmarkTools.jl/blob/master/src/trials.jl#L232).

tkelman:
`%.2f` wastes a fair amount of horizontal space in the table. How often do you think people will use tighter than a single percent here?

PR #9, opened 2016-05-26
Fix show methods in light of JuliaLang/julia#16354

jrevels:


timholy:
:+1: I like this better than using Compat, see https://github.com/JuliaLang/Compat.jl/pull/212.

PR #25, opened 2016-11-02
Import Base.Iterators on v0.6

vchuravy:
Otherwise a deprecation warning is printed at each iteration.

x-ref: https://github.com/JuliaLang/julia/pull/18839


jrevels:
Thanks!

PR #28, opened 2016-11-08
Don't return `nothing` in core benchmarkable function even if output is unassigned

jrevels:
This potentially fixes #6 by reverting the `nothing` return block added in #4. Before this PR:

```julia
julia> @benchmark view(a, 1:2, 1:2) setup = (a = rand(3, 3))
BenchmarkTools.Trial:
  samples:          10000
  evals/sample:     1000
  time tolerance:   5.00%
  memory tolerance: 1.00%
  memory estimate:  0.00 bytes
  allocs estimate:  0
  minimum time:     2.882 ns (0.00% GC)
  median time:      2.892 ns (0.00% GC)
  mean time:        2.901 ns (0.00% GC)
  maximum time:     12.871 ns (0.00% GC)
```

After this PR:

```julia
julia> @benchmark view(a, 1:2, 1:2) setup = (a = rand(3, 3))
BenchmarkTools.Trial:
  samples:          10000
  evals/sample:     998
  time tolerance:   5.00%
  memory tolerance: 1.00%
  memory estimate:  64.00 bytes
  allocs estimate:  1
  minimum time:     17.161 ns (0.00% GC)
  median time:      19.459 ns (0.00% GC)
  mean time:        21.492 ns (7.39% GC)
  maximum time:     725.316 ns (85.27% GC)
```

@yuyichao Was LLVM (maybe Julia?) able to infer that `a` wasn't being modified and `nothing` was being returned, and so it elided the calculation? 

It is possible that the `length(out_vars) == 1` and `else` cases are also vulnerable to this sort of problem; if that's the case, I'm not sure of the right way to fix this issue without breaking the functionality added in #4.

### DIFF
--- a/doc/manual.md
+++ b/doc/manual.md
@@ -10,6 +10,7 @@ Bold links indicate sections that should be read carefully in order to avoid com
     * [Benchmark `Parameters`](#benchmark-parameters)
     * **[Interpolating values into benchmark expressions](#interpolating-values-into-benchmark-expressions)**
     * [Setup and teardown phases](#setup-and-teardown-phases)
+    * [Understanding compiler optimizations](#understanding-compiler-optimizations)
 - [Handling benchmark results](#handling-benchmark-results)
     * [`Trial` and `TrialEstimate`](#trial-and-trialestimate)
     * **[Which estimator should I use?](#which-estimator-should-i-use)**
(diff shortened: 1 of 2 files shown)


jrevels:
For example, this still will elide the `view` allocation:

```
julia> @benchmark (view(a, 1:2, 1:2); b=1) setup=(a = rand(3, 3))
BenchmarkTools.Trial:
  samples:          10000
  evals/sample:     1000
  time tolerance:   5.00%
  memory tolerance: 1.00%
  memory estimate:  0.00 bytes
  allocs estimate:  0
  minimum time:     2.883 ns (0.00% GC)
  median time:      2.894 ns (0.00% GC)
  mean time:        2.917 ns (0.00% GC)
  maximum time:     14.778 ns (0.00% GC)
```

I'm leaning towards documenting this as "if you were to call this function, this allocation would get elided, and that's what BenchmarkTools will report." Does that sound reasonable?

andyferris:
In my experience the compiler is reasonably good at eliding things, but I'd love to know what rules it uses to figure out which function calls are pure (@yuyichao?). I had to go to lengths to force it not to in my (very hacky!) home-made benchmarks. 

(I'm not surprised by the final example - that documentation sounds reasonable IMO because that is what the compiler will do to a "real" function of that structure)

yuyichao:
> Was LLVM (maybe Julia?) able to infer that a wasn't being modified and nothing was being returned, and so it elided the calculation?

I would be great if LLVM can do that for us so we can get it for free ;-p

I believe this is the `alloc_elim_pass!` in type inference in action. Without knowing exactly what code `view` generate, the kind of code it can optimize away are allocation of objects that does not leak and are only read from. It currently doesn't handle assignments too well but that should hopefully be fixed in the future.

Note that this is not the same optimization that uses `@pure` info.

yuyichao:
The old issue for this not being eliminated was https://github.com/JuliaLang/julia/issues/12415. See https://github.com/JuliaLang/julia/pull/16021 for some examples of what it can do and what it can't, including a few other issues linked to it.

jrevels:
Thanks for the info!

Out of curiosity, I'm wondering if the `alloc_elim_pass!` could ellide entire loop bodies in the future? If so, BenchmarkTools is going to have to figure out how to work around that. 

For the example in the OP, BenchmarkTools generates code that looks close to the following:

```julia
@noinline f(a) = view(a, 1:2, 1:2)

function execf()
    a = rand(3, 3) # setup
    evals = 1000
    gc_start = Base.gc_num()
    start_time = time_ns()
    for iter in 1:evals
        f(a)
    end
    sample_time = time_ns() - start_time
    gc_diff = Base.GC_Diff(Base.gc_num(), gc_start)
    time = max(sample_time / evals, 0.001)
    gc_time = max(gc_diff.total_time / evals, 0.0)
    memory = Int(fld(gc_diff.allocd, evals))
    allocs = Int(fld(gc_diff.malloc + gc_diff.realloc +
                     gc_diff.poolalloc + gc_diff.bigalloc,
                     evals))
    return time, gc_time, memory, allocs
end
```

My original hope was that forcibly preventing inlining on `f(a)` would kill a lot of unwanted optimizations. Should I be worried about Julia optimizing away this `for` loop in the future?

andyferris:
@jrevels I have done the trick of
```
@noinline f(x) = func_i_want_to_test(x)
g(x) = (for i = 1:100_000; f(x); end)
@time g(x)
```
with reasonable success.

andyferris:
Thanks @yuyichao for the info! Appreciated : )

yuyichao:
> My original hope was that forcibly preventing inlining on `f(a)` would kill a lot of unwanted optimizations. Should I be worried about Julia optimizing away this for loop in the future?

LOL, this is really feeling like benchmarking in C now. The allocation elim pass should not do this but other optimization might if it decides that `f` is pure so yes, if we starts using the inferred pure in the optimization (this is the optimization @andyferris wants IIRC) `f(a)` might be elided. This can be pretty easily fixed by adding a barrier instruction much like `asm volatile ("" ::: "memory");` in C. so that it forces the function to be inferred as non-pure.

andyferris:
> inferred pure in the optimization (this is the optimization @andyferris wants IIRC) 

Going off topic - the optimization I'm *most* excited about is inference post-inlining, since it allows syntactically some very interesting and powerful API possibilities. But yes, I'm also interested some more powerful constant propagation (e.g. `1+1=2`) and mutables on the stack (roughly #11902).

I think we *could* have `1+1=2` before inferrable pureness with just a few more `@pure` annotations, but anyway, I digress...

andyferris:
> This can be pretty easily fixed by adding a barrier instruction much like asm volatile ("" ::: "memory"); in C. so that it forces the function to be inferred as non-pure.

Volatile is yuck. We could do `@nopure` much like `@noinline` for this case.

PR #29, opened 2016-11-10
Better printing of Trial and thus `@benchmark`

### LINKED ISSUE #13: Print important number prominently
If I understand correctly, the median time & memory estimate are by far the most important numbers coming out of a run. The printing should reflect that.  Maybe:

``` julia
julia> @benchmark sin(1)
BenchmarkTools.Trial:
  median time:      13.00 ns (0.00% GC)
  memory estimate:  0.00 bytes

  minimum time:     13.00 ns (0.00% GC)
  mean time:        13.02 ns (0.00% GC)
  maximum time:     36.00 ns (0.00% GC)
  samples:          10000
  evals/sample:     1000
  time tolerance:   5.00%
  memory tolerance: 1.00%
  allocs estimate:  0
```

mauro3:
Fixes #13

```
julia> @benchmark sin(4)
BenchmarkTools.Trial: 
  median time:      14.856 ns (0.00% GC)
  memory estimate:  0.00 bytes

  minimum time:     13.673 ns (0.00% GC)
  mean time:        15.106 ns (0.00% GC)
  maximum time:     51.258 ns (0.00% GC)
  samples:          10000
  evals/sample:     998
  time tolerance:   5.00%
  memory tolerance: 1.00%
  allocs estimate:  0
```

### DIFF
--- a/doc/manual.md
+++ b/doc/manual.md
@@ -52,16 +52,18 @@ To quickly benchmark a Julia expression, use `@benchmark`:
 ```julia
 julia> @benchmark sin(1)
 BenchmarkTools.Trial:
-  samples:          10000
-  evals/sample:     1000
-  time tolerance:   5.00%
-  memory tolerance: 1.00%
   memory estimate:  0.00 bytes
   allocs estimate:  0
+  --------------
   minimum time:     13.00 ns (0.00% GC)
   median time:      13.00 ns (0.00% GC)
   mean time:        13.02 ns (0.00% GC)
   maximum time:     36.00 ns (0.00% GC)
+  --------------
+  samples:          10000
+  evals/sample:     1000
+  time tolerance:   5.00%
+  memory tolerance: 1.00%
 ```
 
 The `@benchmark` macro is essentially shorthand for defining a benchmark, auto-tuning the benchmark's configuration parameters, and running the benchmark. These three steps can be done explicitly using `@benchmarkable`, `tune!` and `run`:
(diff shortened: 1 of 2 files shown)


jrevels:
*I very strongly do not want to emphasize a specific time estimator over the other estimators*, because the different estimators are useful for different things, and don't want the user to make the mistake of favoring one over the other without cause. If the user did happen to make that mistake, I would prefer them to favor the minimum. 

Thus, my original choice was to list the time estimates in order of increasing value (well, most often - sometimes the median is higher than the mean, though that's very rare since most benchmark timing distributions are right-skewed).

I am okay with separating the memory estimates and the trial metadata from the timing estimates. 

Thus, if you're going to make a change to the Trial pretty printing, I think the format should be the following:

```julia
julia> @benchmark sin(4)
BenchmarkTools.Trial: 
  memory estimate:  0.00 bytes
  allocs estimate:  0
  --------------
  minimum time:     13.673 ns (0.00% GC)
  median time:      14.856 ns (0.00% GC)
  mean time:        15.106 ns (0.00% GC)
  maximum time:     51.258 ns (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     998
  time tolerance:   5.00%
  memory tolerance: 1.00%
```

Also note that the docs have to be update (we should switch to Documenter at some point so we can have actual doctests).

jrevels:
Thanks!

mauro3:
Thank you for the excellent package!

PR #30, opened 2016-11-23
feed quasiquote variables in as arguments

vtjnash:
interpolating these into the AST gives the optimizer too much flexibility
for a benchmark, that can mean the compiler will be able to just constant-fold away the work
this rewrites the quasiquote interpolation variables as setup variables, rather than capturing their values directly in the function, thereby ensuring the compiler can't optimize the function based upon their values

### DIFF
--- a/src/execution.jl
+++ b/src/execution.jl
@@ -173,6 +173,23 @@ function collectvars(ex::Expr, vars::Vector{Symbol} = Symbol[])
     return vars
 end
 
+function quasiquote!(ex::Expr, vars::Vector{Expr})
+    if ex.head === :($)
+        lhs = ex.args[1]
+        rhs = isa(lhs, Symbol) ? gensym(lhs) : gensym()
+        push!(vars, Expr(:(=), rhs, ex))
+        return rhs
+    elseif ex.head !== :quote
+        for i in 1:length(ex.args)
+            arg = ex.args[i]
+            if isa(arg, Expr)
+                ex.args[i] = quasiquote!(arg, vars)
+            end
+        end
+    end
+    return ex
+end
+
 macro benchmark(args...)
     tmp = gensym()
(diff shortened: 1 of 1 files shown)


jrevels:
Thanks! This is a really valuable change, and will only get more useful as the compiler gets smarter.

src/execution.jl
```diff
@@ -203,8 +220,14 @@ macro benchmarkable(args...)
     end
     deleteat!(params, delinds)
 
+    quote_vars = Expr[]
+    core = quasiquote!(core, quote_vars)
```
jrevels: The test failure comes from catching the possible (though extreme) case where `core` here is a `Symbol`.

StefanKarpinski:
Wouldn't it be better to automatically do this for any bindings from the outer scope? I.e. instead of having to quasiquote globals to measure the right thing, just do it automatically.

jrevels:
> Wouldn't it be better to automatically do this for any bindings from the outer scope? I.e. instead of having to quasiquote globals to measure the right thing, just do it automatically.

This was originally the case when this package was still in its initial development phase, but after spending some time with it, I ended up going with the current behavior. It's less magical, less bug-prone, and encourages users to be explicit with regards to (and thus think carefully about) variable scoping in the benchmark expression. I did think about "inverting" the current behavior (e.g. variables would be local by default and global if you "interpolated" them), but that ended up being confusing to use/implement via interpolation syntax.

vtjnash:
Thanks jrevels. I leave it up to you what you want to do with this. With the change you mentioned, it now passes CI (it looks like the mac buildbots had network issues). I just mostly wanted to see if it was possible / how it would work. This is probably not a complete unquote implementation (it doesn't handle nesting, although perhaps it doesn't need to), but I thought it might be interesting to play with.

PR #34, opened 2016-12-29
fix multiline show to follow documented API

stevengj:
This fixes `show` to use the [documented API in 0.6](http://docs.julialang.org/en/latest/manual/types.html#Custom-pretty-printing-1) for deciding on whether to use multiline output.  In particular, the three-argument `show(io, "text/plain", x)` indicates multiline output.   It should still be compatible with 0.4 via `@compat`, since 0.4 used `writemime` for the same effect.

(The `multiline` attribute was removed in JuliaLang/julia#17113, and never worked with IJulia.)


jrevels:
Thanks!

PR #37, opened 2017-01-21
added btime and belapsed macros, fixed prettymemory to use SI units

### LINKED ISSUE #35: provide analogue of `@elapsed`
A lot of times, I just want the minimum time, especially if I'm collecting or comparing a lot of timing measurements.  It would be nice to have an analogue of `@elapsed` that just returns the minimum time, but with all of the `@benchmark` niceties.

I've been using:
```jl
"""
Like `@benchmark`, but returns only the minimum time in ns.
"""
macro benchtime(args...)
    b = Expr(:macrocall, Symbol("@benchmark"), map(esc, args)...)
    :(time(minimum($b)))
end
```

Could something like this be included?

stevengj:
This PR adds two macros, `@btime` and `@belapsed`, that are drop-in replacements for `@time` and `@elapsed` from Base except that they use the `@benchmark` apparatus.  For example:

```jl
julia> @btime sin(1)
  11.410 ns (0 allocations: 0 bytes)
0.8414709848078965
 
julia> @belapsed sin(1)
1.1412412412412412e-8
 ```

It also changes the memory output to use SI prefixes.  Previously, it would output something like `17.34 kb`, which is ambiguous — is a kb 1000 or 1024 bytes (we use 1024)?  The ~~SI~~ ISQ prefix is "KiB" for [kibibyte](https://en.wikipedia.org/wiki/Kibibyte), which unambiguously means 1024 bytes.

Also, it now prints `24 bytes` and not `24.00 bytes` — the decimal point was confusing since (presumably) we will never be able to allocate a fractional number of bytes.

Closes #35.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,3 +1,3 @@
 julia 0.4
-Compat 0.8.0
+Compat 0.9.5
 JLD 0.6.6
--- a/src/execution.jl
+++ b/src/execution.jl
@@ -322,7 +322,6 @@ function generate_benchmark_definition(eval_module, out_vars, setup_vars,
             iters = 2
             while (time() - start_time) < params.seconds && iters ≤ params.samples
                  params.gcsample && BenchmarkTools.gcscrub()
-                 params.gcsample && BenchmarkTools.gcscrub()
                  push!(trial, $(samplefunc)(params)[1:end-1]...)
                  iters += 1
             end
--- a/src/execution.jl
+++ b/src/execution.jl
@@ -28,9 +28,13 @@ end
 sample(b::Benchmark, args...) = error("no execution method defined on type $(typeof(b))")
 _run(b::Benchmark, args...; kwargs...) = error("no execution method defined on type $(typeof(b))")
 
-function Base.run(b::Benchmark, p::Parameters = b.params; kwargs...)
+# return (Trial, result) tuple, where result is the result of the benchmarked expression
+function run_result(b::Benchmark, p::Parameters = b.params; kwargs...)
     return eval(current_module(), :(BenchmarkTools._run($(b), $(p); $(kwargs...))))
 end
+
+Base.run(b::Benchmark, p::Parameters = b.params; kwargs...) =
+    run_result(b, p; kwargs...)[1]
 
 function Base.run(group::BenchmarkGroup, args...; verbose::Bool = false, pad = "", kwargs...)
     result = similar(group)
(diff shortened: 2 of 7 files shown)

### CODE REVIEW

**Comment by stevengj** [comment]
Test failure in 0.6 is due to #38.

File: src/execution.jl [hunk 1]
```diff
@@ -304,15 +315,73 @@ function generate_benchmark_definition(eval_module, out_vars, setup_vars,
...
-                iters += 1
-                iters > params.samples && break
+            params.gcsample && BenchmarkTools.gcscrub()
+            s = $(samplefunc)(params)
+            push!(trial, s[1:end-1]...)
+            return_val = s[end]
+            iters = 2
+            while (time() - start_time) < params.seconds && iters ≤ params.samples
+                 params.gcsample && BenchmarkTools.gcscrub()
+                 params.gcsample && BenchmarkTools.gcscrub()
```
jrevels: Why is this getting done twice here?

stevengj: Typo, sorry

stevengj:
Yay, tests are passing in the rebased branch.

stevengj:
Rebased.

jrevels:
Thanks!

stevengj:
See also JuliaLang/julia#20173

KristofferC:
`@btime` is awesome and something that I have wanted for a long time (without realizing it).

REQUIRE
```diff
@@ -1,3 +1,3 @@
 julia 0.4
-Compat 0.8.0
+Compat 0.9.5
```
tkelman: what is this using from 0.9.5?

stevengj: I can't remember, maybe it was the `Symbol` constructor?  I definitely upgrade the Compat requirement for a reason.

tkelman: `Symbol` is older I believe - probably https://github.com/JuliaLang/Compat.jl/commit/2c11855a34bd2a9e277f5ddbec8eed49df48a02b

Maybe an earlier version of this was using one of the show rewrites? doesn't look like the final version is though

stevengj: I think maybe an earlier version tried to use a buffer and `String(take!(buf))` for the `@btime` test, which requires Compat 0.9.5.   I then rewrote it to use `redirect_stdout(f)` with a temporary file, which requires Compat 0.9.0 (https://github.com/JuliaLang/Compat.jl/commit/2c11855a34bd2a9e277f5ddbec8eed49df48a02b)

PR #38, opened 2017-01-21
fix tests and doc in 0.6: semicolons no longer give macros keywords

stevengj:
This fixes the tests to pass in JuliaLang/julia#7669


jrevels:
Thanks. I merged this PR for the sake of keeping BenchmarkTools forward-compatible with Julia master, but we should remember to add back the semicolon test once https://github.com/JuliaLang/julia/issues/20000 is fixed.

PR #41, opened 2017-01-26
fix macro hygiene problems

### LINKED ISSUE #40: Side effect from btime macro
Why does this change the binding of a? cc @stevengj 

```julia
julia> using BenchmarkTools

julia> a = BitArray(rand() < 0.5 for x in 1:10^4, y in 1:10^4);

julia> @btime sum(a,1)
  52.620 ms (2 allocations: 78.20 KiB)
1×10000 Array{Int64,2}:
 5043  5049  5047  5028  4921  4929  …  5029  5022  5013  5082  5032  5011

julia> a
2
```

stevengj:
This fixes #40 and some other hygiene problems (e.g. the macros were assuming that the local scope did not have a variable named `BenchmarkTools`).

### DIFF
--- a/src/execution.jl
+++ b/src/execution.jl
@@ -195,15 +195,13 @@ function quasiquote!(ex::Expr, vars::Vector{Expr})
 end
 
 macro benchmark(args...)
-    tmp = gensym()
     _, params = prunekwargs(args...)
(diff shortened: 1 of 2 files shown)


stevengj:
Thanks for merging!  Probably a new release should be tagged with the fix.

PR #49, opened 2017-02-27
remove tolerances from all pretty printing but TrialJudgement

jrevels:
...since the tolerances are basically irrelevant unless you're calling `judge`.


jrevels:
Error looks like an unrelated Travis-specific failure.

Issue #21, opened 2016-08-31
Make `warmup` function's `verbose` optional argument a keyword argument for consistency


musm:
Makes sense since  `run`  has this optional keyword argument

jrevels:
[It does have a verbose option](https://github.com/JuliaCI/BenchmarkTools.jl/blob/master/src/execution.jl#L62), but it should be made a keyword argument instead of an optional argument for the sake of consistency.

PR #57, opened 2017-05-12
Make `warmup` `verbose` optional argument

musm:
close https://github.com/JuliaCI/BenchmarkTools.jl/issues/21

### DIFF
--- a/src/execution.jl
+++ b/src/execution.jl
@@ -127,7 +127,7 @@ end
 
 function tune!(b::Benchmark, p::Parameters = b.params;
                verbose::Bool = false, pad = "", kwargs...)
-    warmup(b, false)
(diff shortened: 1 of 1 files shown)


musm:
osx failure is unrelated

src/execution.jl
```diff
@@ -127,7 +127,7 @@ end
 
 function tune!(b::Benchmark, p::Parameters = b.params;
                verbose::Bool = false, pad = "", kwargs...)
-    warmup(b, false)
+    warmup(b; verbose = false)
```
musm: not sure if you prefer the `;` here or `,`

musm:
I don't think we can do anything about the master osx failure been seeing that in several places

PR #60, opened 2017-06-02
allow any Associative type in BenchmarkGroup

### LINKED ISSUE #52: Dict -> OrderedDict in BenchmarkGroup?
I don't know if anyone else feels the same, but it is a little bit annoying to me the fact that when I dispay the results of some benchmarks, they are not shown in insertion order. 
As an example, here for me it is visually difficult to compare the performances of different graph types in [Erdos](https://github.com/CarloLucibello/Erdos.jl) because of the lack of ordering:
```julia
julia> @show res["generators"];
res["generators"] = 16-element BenchmarkTools.BenchmarkGroup:
  tags: []
  ("rrg","Net(500, 750) with [] graph, [] vertex, [] edge properties.") => Trial(632.408 μs)
  ("rrg","Graph{Int64}(100, 150)") => Trial(121.221 μs)
  ("rrg","Net(100, 150) with [] graph, [] vertex, [] edge properties.") => Trial(115.033 μs)
  ("rrg","Graph{Int64}(500, 750)") => Trial(677.647 μs)
  ("complete","Net(100, 4950) with [] graph, [] vertex, [] edge properties.") => Trial(896.223 μs)
  ("complete","DiGraph{Int64}(100, 9900)") => Trial(617.122 μs)
  ("complete","DiNet(20, 380) with [] graph, [] vertex, [] edge properties.") => Trial(42.104 μs)
  ("erdos","Graph{Int64}(500, 1500)") => Trial(405.240 μs)
  ("erdos","Net(100, 300) with [] graph, [] vertex, [] edge properties.") => Trial(71.516 μs)
  ("complete","DiGraph{Int64}(20, 380)") => Trial(23.721 μs)
  ("complete","Net(20, 190) with [] graph, [] vertex, [] edge properties.") => Trial(20.845 μs)
  ("complete","Graph{Int64}(100, 4950)") => Trial(159.900 μs)
  ("complete","DiNet(100, 9900) with [] graph, [] vertex, [] edge properties.") => Trial(1.861 ms)
  ("erdos","Net(500, 1500) with [] graph, [] vertex, [] edge properties.") => Trial(297.167 μs)
  ("complete","Graph{Int64}(20, 190)") => Trial(7.340 μs)
  ("erdos","Graph{Int64}(100, 300)") => Trial(88.091 μs)

```
Would it be reasonable and not too disruptive to use `OrderedDict`s instead of `Dict` in the `BenchmarkGroup` type? 

Yes, I could write down some more appropriate comparison methods, but asking doesn't hurt :)

Cheers,
Carlo

shashi:
closes #52.

### DIFF
--- a/src/groups.jl
+++ b/src/groups.jl
@@ -4,10 +4,10 @@
 
 immutable BenchmarkGroup
     tags::Vector{Any}
-    data::Dict{Any,Any}
+    data::Associative
 end
 
-BenchmarkGroup(tags::Vector, args::Pair...) = BenchmarkGroup(tags, Dict(args...))
+BenchmarkGroup(tags::Vector, args::Pair...) = BenchmarkGroup(tags, Dict{Any,Any}(args...))
 BenchmarkGroup(args::Pair...) = BenchmarkGroup([], args...)
 
 function addgroup!(suite::BenchmarkGroup, id, args...)
--- a/test/GroupsTests.jl
+++ b/test/GroupsTests.jl
@@ -3,6 +3,7 @@
 using Base.Test
 using BenchmarkTools
 using BenchmarkTools: TrialEstimate, Parameters
+using DataStructures
 
 seteq(a, b) = length(a) == length(b) == length(intersect(a, b))
 
@@ -37,12 +38,16 @@ g2["c"] = tc
 
 trial = BenchmarkTools.Trial(Parameters(), [1, 2, 5], [0, 1, 1], 3, 56)
 
-gtrial = BenchmarkGroup([], Dict("t" => trial))
+gtrial = BenchmarkGroup([], "t" => trial)
 
 # tests #
 #-------#
 
 @test BenchmarkGroup() == BenchmarkGroup([], Dict())
+@test BenchmarkGroup([]) == BenchmarkGroup([], Dict())
+@test BenchmarkGroup([], Dict(1=>2)).data == Dict(1=>2)
+@test typeof(BenchmarkGroup([], OrderedDict(1=>1, 2=>1)).data) == OrderedDict{Int,Int}
+@test BenchmarkGroup([], OrderedDict(1=>1, 2=>1)).data == OrderedDict(1=>1,2=>1)
 @test length(g1) == 3
 @test g1["a"] == t1a
 @test g1["b"] == t1b
--- a/test/REQUIRE
+++ b/test/REQUIRE
@@ -1,0 +1,1 @@
+DataStructures


src/groups.jl
```diff
@@ -4,7 +4,15 @@
...
+    data::Associative
+
+    function BenchmarkGroup(tags, data::Dict)
+        new(convert(Vector{Any}, tags), convert(Dict{Any, Any}, data))
+    end
+
+    function BenchmarkGroup(tags, data)
+        new(convert(Vector{Any}, tags), data)
+    end
 end
```
jrevels: If we're allowing `data::Associative`, might as well not have these inner constructors.

src/groups.jl
```diff
@@ -4,7 +4,15 @@
...
 
 BenchmarkGroup(tags::Vector, args::Pair...) = BenchmarkGroup(tags, Dict(args...))
```
jrevels: ...and without the inner constructors, I believe we can still maintain the default behavior by changing `Dict(args...)` here to `Dict{Any,Any}(args...)`.

test/GroupsTests.jl
```diff
@@ -43,6 +43,10 @@ gtrial = BenchmarkGroup([], Dict("t" => trial))
 #-------#
 
 @test BenchmarkGroup() == BenchmarkGroup([], Dict())
+@test BenchmarkGroup([]) == BenchmarkGroup([], Dict())
+@test typeof(BenchmarkGroup([], Dict{Int,Int}(1=>2)).data) == Dict{Any,Any}
+@test BenchmarkGroup([], Dict(1=>2)).data == Dict(1=>2)
+@test BenchmarkGroup([], Base.ImmutableDict(1=>2)).data == Base.ImmutableDict(1=>2)
```
jrevels: Probably worth adding a test for explicitly for using OrderedDicts, since that's the motivation behind the PR.

src/groups.jl
shashi: That seems backwards incompatible -- there are some tests that construct other dicts (e.g. `Dict{String, Trial}` and `Dict{String, TrialEstimate}`) and [some code](https://github.com/JuliaCI/BenchmarkTools.jl/blob/master/src/groups.jl#L49) that assume `.data` is `{Any, Any}`.

test/GroupsTests.jl
shashi: Don't want to add DataStructures as a test dependency, that would again mean more work keeping this package working.

jrevels: I'm pretty conservative about adding non-test dependencies, but test dependencies are fine. I think it's worth doing in this case.

src/groups.jl
jrevels: > there are some tests that construct other dicts (e.g. Dict{String, Trial} and Dict{String, TrialEstimate}) 

If tests are failing just due to over-specialized `BenchmarkGroup` initialization, they should be using the `BenchmarkGroup(::Vector, ::Pair...)` constructor (which should be changed per my comment below).

> and some code that assume .data is {Any, Any}.

That code doesn't assume that `data::Dict{Any, Any}`, just that `typeof(k) <: keytype(dest.data)` and `typeof(f($(getinds...))) <: valtype(dest.data)` (e.g. like `map!` for arrays). There might be other code that actually does assume `Dict{Any,Any}`, though, in which case this PR would have to generalize that code.

This PR is all about allowing the user to select the their own `Associative` type - we shouldn't special case `Dict`s to break that behavior.

shashi: Yeah, I can do that. But it will break other packages using this behavior. Hopefully no one else is.

test/GroupsTests.jl
shashi: This is  a very simple thing storing the Associative you wanted to store in the right field, I think doing it with OrderedDict is an overkill, but okay I can add that dependency.

jrevels:
So here's my idea for preventing the breakage without special-casing `Dict`: 

```julia
immutable BenchmarkGroup
    tags::Vector{Any}
    data::Associative{Any,Any}
    BenchmarkGroup(tags::Vector, data::Associative{Any,Any}) = new(tags, data)
    BenchmarkGroup(tags::Vector, data::Associative) = new(tags, copy!(similar(data, Pair{Any,Any}), data))
end 
```

This way, we still get the main advantage of this PR - being able to use dictionary representations - without causing breakage or sacrificing consistency.

vtjnash:
outdated, but I think we might want to just add a sorted/ordered "keys" Vector?

For the purposes of Nanosoldier, we don't need this, since we want these keys sorted (by repr), not ordered, which we do as a post-processing step

PR #61, opened 2017-06-09
Fix current_module deprecation on 0.7

timholy:


jrevels:
Thanks!

src/execution.jl
```diff
@@ -29,15 +29,15 @@ end
...
+# scope of the module returned by `@__MODULE__`. This is to avoid any weird quirks
 # when calling the experiment from within other contexts.
 
 sample(b::Benchmark, args...) = error("no execution method defined on type $(typeof(b))")
 _run(b::Benchmark, args...; kwargs...) = error("no execution method defined on type $(typeof(b))")
 
 # return (Trial, result) tuple, where result is the result of the benchmarked expression
 function run_result(b::Benchmark, p::Parameters = b.params; kwargs...)
-    return eval(current_module(), :(BenchmarkTools._run($(b), $(p); $(kwargs...))))
+    return eval(@__MODULE__, :(BenchmarkTools._run($(b), $(p); $(kwargs...))))
```
yuyichao: I don't think this is correct. This will always return `BenchmarkTools`. What exactly is the reason to use `current_module()` here before?

jrevels: Ah, good catch - I haven't been following the `@__MODULE__` changes, I had naively assumed it would expand into an equivalent of `current_module()`. 

To be explicit: `current_module` was being used here to grab the module from which this code was being executed. An example of the intended behavior:

```julia
julia> f() = current_module()
f (generic function with 1 method)

julia> module M
           g() = Main.f()
           println("This should print `M`, not `Main`: ", g())
       end;
This should print `M`, not `Main`: M

julia> module M2
           using M
           g2() = M.g()
           println("This should print `M2`, not `M`: ", g2())
       end;
This should print `M2`, not `M`: M2
```

yuyichao: The PR's intent is to remove the dynamic scope semantics so it is not possible to implement `current_module()` anymore. If you have a generic function call, it will be impossible for it to know from where it is called so such information has to be passed down explicitly.

yuyichao: And what I'm also asking is what's the reason these needs to be running in a specific module. Both `lineartrail` and `run_result` seems to be splicing values and not expressions so the module they are evaluated shouldn't matter

timholy:
This turns out to be problematic unless the user has said `using Compat` in Main.

src/execution.jl
```diff
@@ -250,7 +250,7 @@ macro benchmarkable(args...)
 
     # generate the benchmark definition
     return esc(quote
```
yuyichao: The `esc` here is wrong. Not really the fault of the PR.

tkelman:
This should be fixed before tagging.

PR #64, opened 2017-06-29
quell type and immutable deprecation warnings under 0.7

Sacha0:
Wonderful meeting you last week `@jrevels`! :)

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,3 +1,3 @@
-julia 0.4
+julia 0.6
 Compat 0.26
 JLD 0.6.6
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,8 +4,7 @@ os:
   - linux
   - osx
 julia:
-  - 0.4
-  - 0.5
+  - 0.6
   - nightly
 notifications:
   email: false
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,7 +1,7 @@
 environment:
   matrix:
-  - JULIA_URL: "https://julialang-s3.julialang.org/bin/winnt/x86/0.4/julia-0.4-latest-win32.exe"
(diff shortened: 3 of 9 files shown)


Sacha0:
(The CI failures on 0.4 and 0.5 are of course expected, and the macOS nightly failure seems unrelated. Best!)

tkelman:
should raise minimum julia version in REQUIRE to 0.6 then?

Sacha0:
Yes, assuming `@jrevels` is on board?

jrevels:
Sorry for the late response, thanks for this! 

> should raise minimum julia version in REQUIRE to 0.6 then?

Seems like a good idea.

Sacha0:
Done, no worries, my pleasure, and preemptive happy 4th! :)

REQUIRE
```diff
@@ -1,3 +1,3 @@
-julia 0.4
+julia 0.6
```
tkelman: and drop earlier from ci

Sacha0: Thanks! Let's see whether I fixed that correctly...

tkelman: looks right

Sacha0:
Thanks `@jrevels`! :)

tkelman:
I don't think this works correctly when the consumer doesn't use Compat itself. Maybe need to interpolate it into the macro expansion?

Sacha0:
> I don't think this works correctly when the consumer doesn't use Compat itself. Maybe need to interpolate it into the macro expansion?

(Did you mean to post this somewhere else? :) )

tkelman:
No, see the JuliaCIBot failures downstream on the tag pr

tkelman:
ah no whichever pr added at-module calls sorry

PR #66, opened 2017-07-20
add progress meter keyword to run and tune!

KristofferC:
As a good boy I am submitting https://github.com/JuliaCI/PkgBenchmark.jl/pull/38 to where it ideally would be placed but I realize adding an extra dependency is non ideal for this package.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,3 +1,4 @@
 julia 0.6
 Compat 0.26
 JLD 0.6.6
+ProgressMeter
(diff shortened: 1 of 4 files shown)


tkf:
FYI, #153 is another implementation of progress bar.  I think #153 is much more flexible and useful because:

* It let users choose different "progress monitor" appropriate for their needs (e.g., Juno users would see progress bars in the GUI).
* I think the `println`-based `verbose` implementation is better than `showvalues`-based implementation as it leaves timing information for each group. It also is usable in CI services.

vchuravy:
Closed in favour of #153

Issue #69, opened 2017-08-11
broken on master:  invalid syntax


musm:
```julia
julia> f() =0
f (generic function with 1 method)

julia> @btime f()
ERROR: syntax: invalid syntax (escape (call (outerref f)))
Stacktrace:
 [1] generate_benchmark_definition(::Module, ::Array{Symbol,1}, ::Array{Any,1}, ::Expr, ::Void, ::Void, ::BenchmarkTools.Parameters) at C:\Users\Mus\.julia\v0.7\BenchmarkTools\src\execution.jl:289

```

giordano:
I've been seeing this error for a while (more than a week), I was surprised that no one else reported it, so I thought at first that I had something broken on my machine.

I didn't git-bisect Julia to see when this got broken, I just tried two commits:
* on https://github.com/JuliaLang/julia/commit/952dc93489213327147c7d7d14182a81eda35835 `BenchmarkTools` still works
* on https://github.com/JuliaLang/julia/commit/8d85aa47e8dc27f23daf09f0343a7868931ef9ef it's broken

Thus I suspect this is another example of breakage produced by https://github.com/JuliaLang/julia/pull/22985

jrevels:
I'll try to get around to fixing this this week, since it seems to be affecting a lot of people.

In the meantime, if somebody wants to take a crack at this before I get to it, I recommend [investigating the macro scoping bugs pointed out a long time ago](https://github.com/JuliaCI/BenchmarkTools.jl/pull/61#discussion_r121386643), keeping in mind that we need to preserve the behavior I mentioned [here](https://github.com/JuliaCI/BenchmarkTools.jl/issues/65#issuecomment-313926098).

PR #77, opened 2017-09-21
fix eval module scoping and escape issues on v0.7

jrevels:
fixes https://github.com/JuliaCI/BenchmarkTools.jl/issues/69

@yuyichao Can you check this? I'm sure I've messed something up somewhere...


src/execution.jl
```diff
@@ -28,16 +28,15 @@ end
...
+# Note that trials executed via `run` and `lineartrial` are always executed at top-level
+# scope, in order to allow transfer of locally-scoped variables into benchmark scope.
 
 sample(b::Benchmark, args...) = error("no execution method defined on type $(typeof(b))")
 _run(b::Benchmark, args...; kwargs...) = error("no execution method defined on type $(typeof(b))")
 
 # return (Trial, result) tuple, where result is the result of the benchmarked expression
 function run_result(b::Benchmark, p::Parameters = b.params; kwargs...)
-    return eval(@__MODULE__, :(BenchmarkTools._run($(b), $(p); $(kwargs...))))
+    return eval(:($BenchmarkTools._run($(b), $(p); $(kwargs...))))
```
yuyichao: The version before and after change have exactly the same meaning.

src/execution.jl
```diff
@@ -203,12 +202,13 @@ end
...
-        tmp = @benchmarkable $(map(esc,args)...)
-        warmup(tmp)
-        $(hasevals(params) ? :() : :(tune!(tmp)))
-        run(tmp)
-    end
+    tmp = gensym()
+    esc(quote
+        $tmp = $BenchmarkTools.@benchmarkable $(args...)
+        $BenchmarkTools.warmup($tmp)
+        $(hasevals(params) ? :() : :(tune!($tmp)))
```
yuyichao: Also for `tune!`

src/execution.jl
```diff
@@ -286,9 +284,9 @@ function generate_benchmark_definition(eval_module, out_vars, setup_vars,
         invocation = :($(Expr(:tuple, out_vars...)) = $(signature))
         core_body = :($(core); $(returns))
     end
-    eval(eval_module, quote
+    eval(quote
```
yuyichao: So you want to eval this in `BenchmarkTools`?

jrevels: From my understanding, eval location doesn't actually matter, as long as it's at the top-level of some (any) module. That was my thinking after reading [your comment here](https://github.com/JuliaCI/BenchmarkTools.jl/pull/61#discussion_r121555013), anyway - maybe I misunderstood.

yuyichao: My comment wasn't about this case. It's about the two cases above which seems like should be using `invokelatest` instead.
The module you evaluate something in of course has great significance in general and evaluating in the module your code is in is almost always the wrong one (unless it doesn't matter, in which case `eval` is likely not what you want to use, i.e. the case above).
IIUC `core` is user input so it must be evaluated in the module where the user calls this.

jrevels: >The module you evaluate something in of course has great significance in general and evaluating in the module your code is in is almost always the wrong one (unless it doesn't matter, in which case eval is likely not what you want to use, i.e. the case above).

Yeah, I think in this case the evaluation scope doesn't matter, since my only real goal is for it to not match the calling scope. As you mentioned, maybe this indicates that something like `invokelatest` is more appropriate...

> IIUC core is user input so it must be evaluated in the module where the user calls this.

For example,

```julia
julia> for i in 1:10
           @benchmark sin(i)
       end
ERROR: UndefVarError: i not defined
```

...is actually the behavior that I want (and is the current BenchmarkTools behavior). If calling scope matched evaluation scope, I'm worried we might eventually have scope-dependent optimizations that would cause benchmark results to change based on callsite location - which wouldn't be "incorrect", per se, but I think might mislead naive users who aren't aware of what's going on.

yuyichao: What about

```julia
module M
using BenchmarkTools
f(i) = i
@benchmark f(1)
end
```

?

jrevels: Ah, good point, now I feel silly XD

yuyichao: BTW, by "the case above" I am talking about the case that's actually above this one, i.e. the one mentioned in the comment you linked. Those doesn't seem to be accepting (user generated) expressions as inputs.

musm:
bump  : )

mbauman:
Flyby comment: `@btime` is still broken on this branch, but `@benchmark` seems to be working...

jrevels:
Today is an MIT holiday, which means I actually have time to code instead of draft grant proposals! 

All v0.7 depwarns that I've encountered during testing are now resolved, and tests are locally passing on both Julia v0.6 and today's master. I plan on merging and tagging immediately, pending successful CI. 

Special thanks to @yuyichao for his help here.

jrevels:
Tests passing on Linux, but the Mac queue seems backed up..but I think we've waited long enough for this, IMO it's merge time.

ararslan:
I can't thank you enough for taking this on! It's been sad not being able to use `@benchmark` on 0.7.

PR #79, opened 2017-10-20
RFC: Migrate serialization format to JSON

### LINKED ISSUE #15: use more portable/stable serialization/deserialization
While JLD _is_ valid HDF5, I've been told JLD is painfully structured for use in other languages. 

I've been bitten by compatibility issues, as well. [Maintaining backwards compatibility is hard to pull off smoothly](https://github.com/JuliaCI/BenchmarkTools.jl/pull/12), and [JLD changes often conflict with changes in Base, breaking forwards compatibility](https://github.com/JuliaLang/julia/commit/fa5af2377381145c8949c2de7c2bfceaf772d83e#commitcomment-18136477).

It would be nice to have simple CSV or JSON (de)serialization methods that didn't rely on external dependencies (or at least relied on more stable ones). One benefit is that methods like  `BenchmarkTools.save`/`BenchmarkTools.load` would be easier to patch than external methods like `JLD.load`/`JLD.save`.

cc @quinnj, since we discussed this at JuliaCon. Do you have any specific opinions here?

ararslan:
This is probably hot garbage at the moment, but I figured I'd put it up to get feedback while I continue to improve it.

Also note that the dependency on JSON.jl is likely temporary, until I can implement a mini JSON parser in this package.

Fixes #15


src/serialization.jl
```diff
@@ -21,36 +31,113 @@ mutable struct TrialPreV006
...
+        msg = if ext == ".jld"
+            "JLD deserialization is no longer supported. Benchmarks should now be saved\n" *
+            "in JSON format using `save(\"$noext\".json, args...)`. You will need to\n" *
+            "convert existing saved benchmarks to JSON in order to deserialize them with\n" *
+            "this version of BenchmarkTools."
+        else
+            "Only JSON deserialization is supported. Use `load(\"$noext.json\", args...)`."
+        end
+        throw(ArgumentError(msg))
+    end
```
jrevels: Might want to break out this duplicated assert into its own function (either rewriting the message to be the same for load/save or just pass a flag giving the context).

src/serialization.jl
```diff
@@ -21,36 +31,113 @@ mutable struct TrialPreV006
...
+    if !isempty(args)
+        throw(ArgumentError("Looking up deserialized values by name is no longer supported, " *
+                            "as names are no longer saved."))
     end
-    return nothing
+    parsed = open(JSON.parse, filename, "r")
+    if !isa(parsed, Vector) || length(parsed) != 2 || !isa(parsed[1], Dict) || !isa(parsed[2], Vector)
+        error("Unexpected JSON format. Was this file produced by BenchmarkTools?")
```
jrevels: "produced by" --> "originally written by" would be clearer IMO

src/serialization.jl
```diff
@@ -1,6 +1,16 @@
-const VERSION_KEY = "__versions__"
+const VERSIONS = Dict("Julia" => string(VERSION),
+                      "BenchmarkTools" => string(BENCHMARKTOOLS_VERSION))
 
-const VERSIONS = Dict("Julia" => string(VERSION), "BenchmarkTools" => string(BENCHMARKTOOLS_VERSION))
+# TODO: Add any new types as they're added
+const supported_types = [Benchmark, BenchmarkGroup, Parameters, TagFilter, Trial,
```
jrevels: Could we change this to `SUPPORTED_TYPES`?

test/SerializationTests.jl
```diff
@@ -1,14 +1,68 @@
...
+    @test_throws ArgumentError BenchmarkTools.load("x.jld")
+    @test_throws ArgumentError BenchmarkTools.load("x.txt")
+    @test_throws ArgumentError BenchmarkTools.load("x.json", "b")
+end
+
+@testset "Error checking" begin
+    isfile(tmp) && rm(tmp)
+    open(tmp, "w") do f
+        print(f, """
+        {"never":1,[{"gonna":2,"give":3,"you":4,"up":5}]}
```
jrevels: Please consult [the manual](https://www.youtube.com/watch?v=dQw4w9WgXcQ) before adding a test like this.

src/serialization.jl
```diff
@@ -21,36 +31,113 @@ mutable struct TrialPreV006
...
+#        result = JLD.load(filename, args...)
+#    else
+#        JLD.translate("BenchmarkTools.Parameters", "BenchmarkTools.ParametersPreV006")
+#        JLD.translate("BenchmarkTools.Trial", "BenchmarkTools.TrialPreV006")
+#        result = JLD.load(filename, args...)
+#        JLD.translate("BenchmarkTools.Parameters", "BenchmarkTools.Parameters")
+#        JLD.translate("BenchmarkTools.Trial", "BenchmarkTools.Trial")
+#    end
+#    return result
+#end
```
jrevels: I'm sure you'd do this eventually anyway, but just to be explicit, we should definitely delete this commented out stuff before merging...what a joy it will be to be rid of it!

test/SerializationTests.jl
KristofferC: !!!!!!!!

src/serialization.jl
ararslan: Yeah definitely

test/SerializationTests.jl
ararslan: You're right Jarrett, my apologies. We're no strangers to unit testing; you know the rules and so do I.

ararslan:
I'm not sure what to make of that error message. I'm not seeing it locally. :/

ararslan:
TESTS ARE PASSING ON 0.6

ararslan:
I've verified locally that the new serialization tests pass on 0.7. There are other issues, likely due to the 0.7 macro hygiene chaos, which are keeping the full test suite from passing.

src/serialization.jl
```diff
@@ -1,56 +1,87 @@
...
 end
 
-function JLD.readas(p::ParametersPreV006)
-    return Parameters(p.seconds, p.samples, p.evals, Float64(p.overhead), p.gctrial,
-                      p.gcsample, p.time_tolerance, p.memory_tolerance)
+function recover(x::Vector)
+    length(x) == 2 || throw(ArgumentError("Expecting a vector of length 2"))
+    typename = x[1]::String
+    fields = x[2]::Dict
+    T = eval(parse(typename))::Type
```
KristofferC: Isn't this a bit scary to call on something that comes from a json file? Should there be a whitelist of types and just switch on that?

ararslan: Yeah I suppose so. The hope is that people will only read in JSON files that were produced by this package, but I suppose it's pretty easy to falsify that if you know the expected format.

ararslan: Apparently Jarrett is unfazed by this.

KristofferC: If it is valid JSON syntax, I guess it should be fine?

ararslan:
I tried adding this to address @KristofferC's comment but for some reason it fails with an `UndefVarError` for `NefariousCode` rather than the expected failure...

```diff
diff --git a/src/serialization.jl b/src/serialization.jl
index 52715ed..ba3e9e1 100644
--- a/src/serialization.jl
+++ b/src/serialization.jl
@@ -22,6 +22,10 @@ function recover(x::Vector)
     length(x) == 2 || throw(ArgumentError("Expecting a vector of length 2"))
     typename = x[1]::String
     fields = x[2]::Dict
+    if !any(t->ismatch(Regex(string('^', t, "({[^}]+)?\$")), typename), SUPPORTED_TYPES)
+        throw(ArgumentError("Cannot deserialize type '$typename'; only BenchmarkTools " *
+                            "types can be deserialized"))
+    end
     T = eval(parse(typename))::Type
     fc = fieldcount(T)
     xs = Vector{Any}(fc)
diff --git a/test/SerializationTests.jl b/test/SerializationTests.jl
index 7a51ae7..54f1f8f 100644
--- a/test/SerializationTests.jl
+++ b/test/SerializationTests.jl
@@ -78,6 +78,22 @@ end
         end
     end

+    withtempdir() do
+        tmp = joinpath(pwd(), "tmp.json")
+        open(tmp, "w") do f
+            print(f, """
+            [{"Julia":"1.2.3","BenchmarkTools":"3.2.1"},[["NefariousCode",{}]]]
+            """)
+        end
+        try
+            BenchmarkTools.load(tmp)
+            error("aaahhh")
+        catch err
+            @test err isa ArgumentError
+            @test contains(err.msg, "Cannot deserialize type 'NefariousCode'")
+        end
+    end
+
     @test_throws ArgumentError BenchmarkTools.recover([1])
 end
```

KristofferC:
I think you are missing an end brace in your regex, so should be `"({[^}]+})?\$")`. However, you could just use a type like `BenchmarkTools.Benchmark{println(1+1)}"`.

On the other hand, I think jld files can already execute arbitrary code so I guess this isn't worse than before.

jrevels:
LGTM! Some of these removed lines were a long time coming.

PR #82, opened 2017-10-24
add quick start to README

stevengj:
The vast majority of users probably just want `@btime`, so I think it is useful to include a "quick start" section in the manual that just gives a couple of examples (including interpolation).

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,6 +17,35 @@ To install BenchmarkTools, you can run the following:
 Pkg.add("BenchmarkTools")
 ```
 
+## Quick Start
+
+The simplest usage is via the `@btime` macro, which is analogous to Julia's built-in [`@time` macro](https://docs.julialang.org/en/stable/stdlib/base/#Base.@time) but is often more accurate (by collecting results over multiple runs):
+
+```julia
+julia> using BenchmarkTools, Compat   # you need to use both modules
+
+julia> @btime sin(1)
+  15.081 ns (0 allocations: 0 bytes)
+0.8414709848078965
+```
+
+If the expression you want to benchmark depends on global variables, you should use `$` to "interpolate" them into the benchmark expression to [avoid the problems of benchmarking with globals](https://docs.julialang.org/en/stable/manual/performance-tips/#Avoid-global-variables-1).  Essentially, any interpolated variable `$x` or expression `$(...)` is "pre-computed" before benchmarking begins:
+
+```julia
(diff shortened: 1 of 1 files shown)


ararslan:
I always assumed that most users would want `@benchmark` rather than `@btime`.

stevengj:
I think `@benchmark` gives too much information 99% of the time.

ararslan:
Okay, makes sense.

README.md
```diff
@@ -17,6 +17,35 @@ To install BenchmarkTools, you can run the following:
...
+
+```julia
+julia> using BenchmarkTools, Compat   # you need to use both modules
+
+julia> @btime sin(1)
+  15.081 ns (0 allocations: 0 bytes)
+0.8414709848078965
+```
+
+If the expression you want to benchmark depends on global variables, you should use `$` to "interpolate" them into the benchmark expression to [avoid the problems of benchmarking with globals](https://docs.julialang.org/en/stable/manual/performance-tips/#Avoid-global-variables-1).  Essentially, any interpolated variable `$x` or expression `$(...)` is "pre-computed" before benchmarking begins:
```
jrevels: > If the expression you want to benchmark depends on global variables

Might be more clear for naive users if we change `global` to `external` here. 

Also might be a good idea to link to https://github.com/JuliaCI/BenchmarkTools.jl/blob/master/doc/manual.md#interpolating-values-into-benchmark-expressions from here.

README.md
```diff
@@ -17,6 +17,35 @@ To install BenchmarkTools, you can run the following:
 Pkg.add("BenchmarkTools")
 ```
 
+## Quick Start
```
jrevels: Can we move this section below the `Documentation` section?

stevengj: The whole point is that this is something people will read *before* looking at the full manual...

README.md
stevengj: Okay, will do.

README.md
jrevels: The `Documentation` section is just a couple of links to the manual and reference guide, and it's quite normal style to include such links at the top of a README (along with test/CI badges, etc). This way, users can immediately see what documentation is available without having to dig through other documentation to get there, and it doesn't negatively impact the usefulness of the quick start guide in any way.

stevengj: Okay.

jrevels:
Thanks again!

PR #90, opened 2017-12-01
fix deprecated array constructor

fredrikekre:


fredrikekre:
needs https://github.com/JuliaLang/METADATA.jl/pull/12295

ararslan:
For whatever reason, I made the module name `IterativeEigenSolvers` (note the uppercase S), so the `using` will need to be adjusted.

fredrikekre:
Ah, thanks! I got it right in the commit message at least :)

fredrikekre:
Fixed. Btw, `IterativeEigenSolvers` is only needed for this: https://github.com/JuliaCI/BenchmarkTools.jl/blob/c684b3e5c8d347a1175c8b349a4f5bede18681d9/test/ExecutionTests.jl#L27 so could perhaps be worth replacing that kwarg test with something else and not depend on `IterativeEigenSolvers`

PR #94, opened 2018-01-18
[RFC/WIP] Measure cpu-time and real-time and report both

vchuravy:
This still needs some cleanup and the history is a bit messy.
In essence this measures both the time actually spend running on the cpu
and the time passed. If the OS migrates a process the cpu-time won't increase,
while real-time will. Similar for other things like calling into the kernel.

- [ ] Update docs


src/execution.jl
```diff
@@ -314,21 +314,24 @@ function generate_benchmark_definition(eval_module, out_vars, setup_vars, core,
             $(setup)
             __evals = __params.evals
             __gc_start = Base.gc_num()
-            __start_time = time_ns()
+            __start_realtime = BenchmarkTools.Timers.realtime()
+            __start_cputime = BenchmarkTools.Timers.cputime()
```
jrevels: Interpolate `BenchmarkTools` here.

Also, this is awesome!

src/timers/darwin.jl
```diff
@@ -0,0 +1,43 @@
+const RUSAGE_SELF = Cint(0)
+const RUSAGE_CHILDREN = Cint(-1)
+
+struct timeval
```
ararslan: This and `rusage` should use the standard Julia naming convention of CamelCase for types. So this should be `TimeVal` and the other should be `RUsage`.

src/trials.jl
```diff
@@ -4,53 +4,59 @@
...
+    push!(t.cputimes, cputime)
     push!(t.gctimes, gctime)
     memory < t.memory && (t.memory = memory)
     allocs < t.allocs && (t.allocs = allocs)
     return t
 end
 
 function Base.deleteat!(t::Trial, i)
-    deleteat!(t.times, i)
+    deleteat!(t.realtimes, i)
```
vchuravy: Missing `deleteat(t.cputimes)`

src/execution.jl
```diff
@@ -409,7 +412,7 @@ macro btime(args...)
         $trialmin = $BenchmarkTools.minimum($trial)
         $trialallocs = $BenchmarkTools.allocs($trialmin)
         println("  ",
-                $BenchmarkTools.prettytime($BenchmarkTools.time($trialmin)),
+                $BenchmarkTools.prettytime($BenchmarkTools.realtime($trialmin)),
```
vchuravy: Should we print `cputime` here as well? or `cpuratio`?

src/groups.jl
```diff
@@ -81,7 +81,10 @@ Base.median(group::BenchmarkGroup) = mapvals(median, group)
 Base.min(groups::BenchmarkGroup...) = mapvals(min, groups...)
 Base.max(groups::BenchmarkGroup...) = mapvals(max, groups...)
 
-Base.time(group::BenchmarkGroup) = mapvals(time, group)
+import Base: time
+@deprecate time(group::BenchmarkGroup) realtime(group)
```
vchuravy: The `@deprecate` should be moved to a central location

src/trials.jl
```diff
@@ -4,53 +4,59 @@
 
 mutable struct Trial
     params::Parameters
-    times::Vector{Float64}
+    realtimes::Vector{Float64} #TODO: rename to realtimes
```
vchuravy: remove todo

src/timers/windows.jl
```diff
@@ -0,0 +1,36 @@
...
+@inline function cputime()
+    proc = ccall(:GetCurrentProcess, HANDLE, ())
+    creation_time = Ref{FileTime}()
+    exit_time = Ref{FileTime}()
+    kernel_time = Ref{FileTime}()
+    user_time = Ref{FileTime}()
+
+    err = ccall(:GetProccessTimes, Cint, (HANDLE, Ref{FileTime}, Ref{FileTime}, Ref{FileTime}, Ref{FileTime}), 
+                                proc, creation_time, exit_time, kernel_time, user_time)
+    # TODO: Error handling
```
vchuravy: add error handling

src/timers/windows.jl
```diff
@@ -0,0 +1,36 @@
...
+const HANDLE = Ptr{Void}
+
+@inline function maketime(kernel_time::FileTime, user_time::FileTime)
+    kernel = (kernel_time.dwHighDateTime % UInt64) << 32 | kernel_time.dwLowDateTime
+    user   = (  user_time.dwHighDateTime % UInt64) << 32 |   user_time.dwLowDateTime
+    (kernel + user) * 1e2
+end
+
+@inline function realtime()
+    # counter = Ref{UInt64}()
```
vchuravy: delete dead code

src/timers/darwin.jl
```diff
@@ -0,0 +1,43 @@
...
+end
+
+@inline function realtime()
+    Float64(Base.time_ns())
+end
+
+@inline function cputime()
+    ru = Ref{rusage}()
+    err = ccall(:getrusage, Cint, (Cint, Ref{rusage}), RUSAGE_SELF, ru)
+    # TODO error check
```
vchuravy: add error handling

src/trials.jl
```diff
@@ -4,53 +4,59 @@
...
     allocs::Int
 end
 
-Trial(params::Parameters) = Trial(params, Float64[], Float64[], typemax(Int), typemax(Int))
+Trial(params::Parameters) = Trial(params, Float64[], Float64[], Float64[], typemax(Int), typemax(Int))
 
 @compat function Base.:(==)(a::Trial, b::Trial)
     return a.params == b.params &&
-           a.times == b.times &&
+           a.realtimes == b.realtimes &&
```
vchuravy: Missing: `a.cputimes == b.cputimes`

src/trials.jl
```diff
@@ -168,37 +184,44 @@ function ratio(a::TrialEstimate, b::TrialEstimate)
...
+    realtime::Symbol
+    cputime::Symbol
     memory::Symbol
 end
 
 function TrialJudgement(r::TrialRatio)
     ttol = params(r).time_tolerance
     mtol = params(r).memory_tolerance
-    return TrialJudgement(r, judge(time(r), ttol), judge(memory(r), mtol))
+    return TrialJudgement(r, judge(realtime(r), ttol), judge(cputime(r), ttol), judge(memory(r), mtol))
```
vchuravy: @jrevels I think using the same tolerance for `cputime` and `realtime` should be ok, right?

src/BenchmarkTools.jl
```diff
@@ -10,7 +10,13 @@ if VERSION >= v"0.7.0-DEV.3052"
     using Printf
 end
 
-const BENCHMARKTOOLS_VERSION = v"0.2.2"
+const BENCHMARKTOOLS_VERSION = v"0.3.0"
```
vchuravy: @ararslan Note that I think this is a big enough change the we should bump the minor number, but I can remove this from this PR.

src/trials.jl
```diff
@@ -262,10 +285,10 @@ function prettymemory(b)
     return string(@sprintf("%.2f", value), " ", units)
 end
 
-Base.show(io::IO, t::Trial) = print(io, "Trial(", prettytime(time(t)), ")")
-Base.show(io::IO, t::TrialEstimate) = print(io, "TrialEstimate(", prettytime(time(t)), ")")
-Base.show(io::IO, t::TrialRatio) = print(io, "TrialRatio(", prettypercent(time(t)), ")")
-Base.show(io::IO, t::TrialJudgement) = print(io, "TrialJudgement(", prettydiff(time(ratio(t))), " => ", time(t), ")")
+Base.show(io::IO, t::Trial) = print(io, "Trial(", prettytime(realtime(t)), ")")
```
vchuravy: Should these `show` methods include `cputime`?

src/trials.jl
jrevels: Sure, the tolerance is a fudge number anyway.

src/execution.jl
jrevels: `cpuratio` seems like a good idea

src/trials.jl
jrevels: maybe `cpuratio`

vchuravy:
The code in `timers` is mostly taken from `google/benchmark` and that is under Apache v2.0. The code is fairly basic, but in order to make sure I retained the copyright notice as mandated by the Apache License.
So the whole `Timers` module is under Apache v2

src/BenchmarkTools.jl
ararslan: I think it's fine to have in this PR. Even though this is recorded in the JSON stuff used by Nanosoldier, the version isn't actually used by anything AFAIK. Thanks for the heads up!

ararslan:
I don't think the Apache license mandates that derived works be licensed under Apache, as I understand it it's like MIT in that you just need to preserve the original license when distributing the original code (in whole or in part). So in this case I think it's worth acknowledging that the code is derived from Google stuff but I don't think it requires a separate license.

vchuravy:
The Apache license does mandate that derived works keep the license notice intact and include a copy of the license and mention all modifications...

> "Derivative Works" shall mean any work, whether in Source or Object form, that is based on (or derived from) the Work and for which the editorial revisions, annotations, elaborations, or other modifications represent, as a whole, an original work of authorship. For the purposes of this License, Derivative Works shall not include works that remain separable from, or merely link (or bind by name) to the interfaces of, the Work and Derivative Works thereof.

> a; You must give any other recipients of the Work or Derivative Works a copy of this License; and
> b; You must cause any modified files to carry prominent notices stating that You changed the files; and
> c; You must retain, in the Source form of any Derivative Works that You distribute, all copyright, patent, trademark, and attribution notices from the Source form of the Work, excluding those notices that do not pertain to any part of the Derivative Works; and

So yeah, I would go for the conservative option and consider a translation of code a derivative work.

ararslan:
Right, we just need to include or link to their license, we don't have to license our own under Apache.

src/trials.jl
vchuravy: On playing with it I think it is not necessary.

vchuravy:
That is contrarian to my understanding of the license.

I translated code under the Apache license to Julia, therefore creating a derivative work. According to the license I must a; keep the license notice, b; include a prominent notice that I changed the files and c; I must give any recipient a copy of the license.

So the code under `src/timers` is Apache licensed and everything else is MIT.

ararslan:
It's not a copyleft license; derivative works don't need to bear the same license. See https://www.whitesourcesoftware.com/whitesource-blog/top-10-apache-license-questions-answered/.

vchuravy:
Ok, changed the license to MIT, while complying with the Apache requirement of retaining the license notices.
This is pretty much ready to merge from my side.

vchuravy:
I would appreciate if people on arcane platforms like Windows or Mac could give this I try. This should work on more reasonable platforms like a current FreeBSD.
But I personally have only tested this on Linux 64bit.

src/timers/windows.jl
```diff
@@ -0,0 +1,43 @@
...
+end
+
+@inline function cputime()
+    proc = ccall(:GetCurrentProcess, HANDLE, ())
+    creation_time = Ref{FileTime}()
+    exit_time = Ref{FileTime}()
+    kernel_time = Ref{FileTime}()
+    user_time = Ref{FileTime}()
+
+    ccall(:GetProccessTimes, Cint, (HANDLE, Ref{FileTime}, Ref{FileTime}, Ref{FileTime}, Ref{FileTime}),
```
amellnik: Typo in Process (found by @ararslan)

vchuravy:
Let's hold merging this for now. I got a report that this reports odd things on Windows.

doc/manual.md
```diff
@@ -44,6 +44,20 @@ Before we get too far, let's define some of the terminology used in this documen
...
 recording the total time `t` it takes to record `n` evaluations, and estimating the sample's time per evaluation as `t/n`. For example, if a sample takes 1 second for 1 million evaluations, the approximate time per evaluation for that sample is 1 microsecond. It's not obvious what the right number of evaluations per sample should be for any given benchmark, so BenchmarkTools provides a mechanism (the `tune!` method) to automatically figure it out for you.
 
+## The difference between real-time and cpu-time
+BenchmarkTools measures both real-time and cpu-time. Real-time is the time that has passed while running your program and cpu-time is the time
+that a processor has spend executing your program. As an example when the operating system deschedules your program, real-time is still passing,
+but cpu-time is paused, this is also the case when your program spends time in the kernel or is blocked waiting. As an example take a program that
```
vchuravy: This is not quite correct. Since cpu-time will increase with kernel work, but will not increase if the kernel puts the process to sleep.

ararslan:
Confirmed that tests pass on current Julia master on FreeBSD 11.1.

doc/manual.md
```diff
@@ -96,11 +110,11 @@ returns the minimum time in seconds.
 
 ```julia
 julia> @btime sin(1)
-  13.612 ns (0 allocations: 0 bytes)
```
KristofferC: I wonder if the GC percentage is worth showing when there are 0 allocations

vchuravy:
Ok hunted down the Windows issue and updated the docs. It turns out the accuracy of the windows counters is only 100ns and so the information becomes fairly useless for short functions.
Jarrett and I discussed a variety of possibilities to deal with this, but in the end we decided to simply turn of printing for cpu-time on Windows.

vchuravy:
I was informed that I missed https://msdn.microsoft.com/en-us/library/windows/desktop/ms684929(v=vs.85).aspx (https://msdn.microsoft.com/en-us/library/windows/desktop/dn553408(v=vs.85).aspx) and so there is hope that we get accurate timings on windows after all.

vchuravy:
So I am currently blocked on converting the cycles to a reasonable time.

- `tsc_invariant`\`tsc_constant` are necessary http://oliveryang.net/2015/09/pitfalls-of-TSC-usage/
   `tsc_invariant` is on Nehalem-and-later
- Getting the TSC frequency is tricky.
- Reading `MSR_PLATFORM_INFO [15:8]` through the `rdmsr` instruction, which can't be executed in user-space.
- Reading through the Intel Manual I noticed that there is a section for cpuid `Time Stamp Counter and Nominal Core Crystal Clock Information Leaf` which has the necessary information sometimes...
 
```
function cpuid(infoType)
         info = Ref{NTuple{4, UInt32}}()
         ccall(:jl_cpuid, Cvoid, (Ref{NTuple{4, UInt32}}, UInt32), info, infoType)
         info[]
end
tsc_invariant() = cpuid(0x80000007)[4] == 0x00000100
function tsc_freq()
          info = cpuid(0x15)
          if info[1] == 0 || info[3] == 0
            return nothing
          else
             return info[2] / info[1] * info[3]
          end
end
```

and on my machine `info[3] == 0`
The manual specifies:
```
TSC_Value = (ART_Value * CPUID.15H:EBX[31:0] )/ CPUID.15H:EAX[31:0] + K
```

jrevels:
> So I am currently blocked on converting the cycles to a reasonable time.

We can always merge this without the Windows-specific improvements (and just print `N/A` or w/e for cpu-time on Windows), and include the improvements in a future PR once they're ready.

vchuravy:
I have to double check that the frequency of `QueryProcessCycleTime` and `QueryPerformanceCounter` are the same, but otherwise I found a workable solution for Windows.

I will let this stew for a bit and focus other work.

src/timers/windows.jl
```diff
@@ -41,3 +41,52 @@ end
...
+    return counter[]
+end
+
+struct Measurement
+    time::UInt64
+    cpu::UInt64
+    function Measurement()
+        proc = currentprocess()
+        time = perfcounter()
+        cpu = cpucycles()
```
vchuravy: `cpucyles(proc)`

goretkin:
This feature is very exciting. Is there any technical issue that is preventing this PR from being merged?

It seems even as far as CPU Time goes, there may be multiple measures in certain OSes (e.g. https://linux.die.net/man/2/getrusage mentions RUSAGE_THREAD) for measuring per-thread times, too, which might suggest there's room for more features in future PRs.

vchuravy:
> This feature is very exciting. Is there any technical issue that is preventing this PR from being merged?

I think I got annoyed at Windows, but I don't recall the details anymore.

gdalle:
Hey @vchuravy, you made me a maintainer, now you deal with the fallout 😈 
This is a pretty old PR with a lot of unresolved conversations, how mergeable do you think it is (provided we can update the branch)?
If the Windows version is not ready, I agree with what @jrevels said:
> We can always merge this without the Windows-specific improvements (and just print N/A or w/e for cpu-time on Windows), and include the improvements in a future PR once they're ready.

vchuravy:
I think having something like this would still be valuable, over the years I have been debating what the best level of integrations might be, experimenting with LIKWID.jl and LinuxPerf.jl (@crstnbr) might have opinions as well.

You can see this PR as an item on my to-do list (but its pretty much at the bottom of the priority queue), or as an invitation and initial stub for someone else to implement it.

willow-ahrens:
What's the difference between Timers and https://github.com/schmrlng/CPUTime.jl ?

willow-ahrens:
@Zentrik what do you think of this PR?

vchuravy:
> What's the difference between Timers and https://github.com/schmrlng/CPUTime.jl ?

The code here precedes that package.

willow-ahrens:
Would there be a maintainability difference? Should we contribute this code to that package?

giordano:
> I think having something like this would still be valuable

For what is worth, these days I've been looking into bad scaling observed on certain system and using code from this PR to measure cpu-time vs real-time was very useful to gain more insights about was going on.

willow-ahrens:
Okay. I'm really excited about merging a PR to do both CPU time as a default, and also optionallly LinuxPerf. I want to make sure that any defaults are cross-platform, so I'll make sure to make that happen. I am planning to revamp the measuring infrastructure just a little bit to make it more modular to different kinds of experiments, and after that I'll revisit this and probably merge it after a few tweaks.

PR #112, opened 2018-07-03
Julia0.7 fixes

slmcbane:
Small PR, I just fixed depwarns on use of `mean` and `median` by using `Statistics` where appropriate. Two other changes:

- Newer versions of Compat remove the IterativeEigensolvers module. I didn't know what you'd want to do with this test; I made it conditional on Arpack.jl being installed.
- I haven't done any package development of my own, so I don't know if my updates to the CI script and REQUIRE are quite correct.

### DIFF
--- a/test/ExecutionTests.jl
+++ b/test/ExecutionTests.jl
@@ -3,8 +3,7 @@ module ExecutionTests
 using BenchmarkTools
 using Compat
 using Compat.Test
-
-using Compat.IterativeEigensolvers
+using Pkg
 
 seteq(a, b) = length(a) == length(b) == length(intersect(a, b))
 
@@ -26,7 +25,9 @@ for s in sizes
 end
 
 groups["special"]["macro"] = @benchmarkable @test(1 == 1)
-groups["special"]["kwargs"] = @benchmarkable svds(rand(2, 2), nsv = 1)
+if "Arpack" in keys(Pkg.installed())
+    groups["special"]["kwargs"] = @benchmarkable svds(rand(2, 2), nsv = 1)
+end
 groups["special"]["nothing"] = @benchmarkable nothing
 groups["special"]["block"] = @benchmarkable begin rand(3) end
 groups["special"]["comprehension"] = @benchmarkable [s^2 for s in sizes]
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,6 @@ os:
   - linux
   - osx
 julia:
-  - 0.6
   - nightly
 notifications:
   email: false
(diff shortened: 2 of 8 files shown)


ararslan:
Note that in the currently released 0.7-beta, the Statistics stdlib module is not a thing. Thus I think a simpler solution would be to keep 0.6 and just do
```julia
if VERSION >= v"0.7.0-beta.85"
    import Statistics: mean, median
else
    import Base: mean, median
end
```
in `src/BenchmarkTools.jl` and remove the module qualification when overloading those functions. Then add
```julia
if VERSION >= v"0.7.0-beta.85"
    using Statistics
end
```
in `test/runtests.jl`. It doesn't have to be in all of the test files.

slmcbane:
Oops, didn't realize about the beta since I've been using nightly for my 0.7 needs.

I think it does need to be in all of the test files because the tests are encapsulated in modules.

test/ExecutionTests.jl
```diff
@@ -26,7 +25,9 @@ for s in sizes
 end
 
 groups["special"]["macro"] = @benchmarkable @test(1 == 1)
-groups["special"]["kwargs"] = @benchmarkable svds(rand(2, 2), nsv = 1)
+if "Arpack" in keys(Pkg.installed())
```
ararslan: I think we should just change the code being benchmarked to not use `svds`.

slmcbane: That sounds good to me.

ararslan:
Nice work, looks good to me now. Would you be willing to add 0.7 to the Travis matrix as well? Just specify 0.7 as another line between 0.6 and nightly. (Note that 0.7 and nightly don't refer to the same version.)

slmcbane:
Sure thing. I'm just getting my feet wet with CI testing, not sure what established practices are.

ararslan:
The Travis and AppVeyor configuration files floating around in various packages provide a reasonable example of how to do things. They all look basically the same, so any you find for a package that's been updated any time in the last year or so should be okay.

On Travis, since Julia is available as a preinstalled language, we can specify a matrix by version for testing. `0.6` uses the latest 0.6.x release (currently 0.6.3), ditto for `0.7` (currently 0.7.0-beta), and `nightly` pulls the latest nightly version. Aside from that, we can let it use the default testing script.

Things are a little different for AppVeyor: it doesn't have Julia by default, so you have to give explicit URLs for downloading, and it doesn't have a default script for Julia. But really testing on AppVeyor is only useful if your package is ecosystem-critical (i.e. a lot of packages depend on it, like StatsBase), it contains any platform-specific logic, or if it has binary dependencies that support Windows.

PR #115, opened 2018-08-16
Update manual.md

adamslc:
Update `replace` to use a pair, and add a semicolon to the end of `teststr`.


adamslc:
When I copy-pasted that block of code into the repl, the output from `teststr` filled several pages of my screen, which was mildly annoying. I expect that most people who are executing that code don't actually care about the details of the benchmarks; they just want a toy suite to play around with.

PR #116, opened 2018-08-23
Make `TrialEstimate` variable naming consistent with their string names in `show(..., t::Trial)`

thchr:
I was playing around with the `Base.show(io::IO, ::MIME"text/plain", t::Trial)` method and got caught by this double-flipping of the meanings of `med`, `max`, & `avg` vs. `medstr`, `maxstr`, & `meanstr`. 

This change doesn't actually change anything output-wise; it merely makes it less painful to read the code as a human.

Finally, it would make decent sense - to me at least - if `meanstr` was swapped for `avgstr` (for consistency with `avg`); I can make that change in here as well, if desired.


jrevels:
Woah, nice catch. That was quite a silly typo!

Issue #96, opened 2018-01-22
Update docs on what to use instead of showall


mauro3:
`showall` is gone: https://github.com/JuliaLang/julia/pull/22847 How to best look at results now? Note that `repr` and `print` show even less info.

devmotion:
I was wondering the same and noticed that `show(stdout, MIME"text/plain"(), suite)` displays only a limited and non-verbose summary. For a more verbose output one has to provide the `verbose` and `limit` keywords, see https://github.com/JuliaCI/BenchmarkTools.jl/blob/master/src/groups.jl#L253). In my opinion it would be nice if BenchmarkTools would respect common properties of `IOContext` such as `limit` and `compact`.

jrevels:
Yup, BenchmarkTools was originally written before we had the nice `IOContext` interface. It'd be great for BenchmarkTools to be updated to properly respect it.

PR #121, opened 2018-10-16
something closer to new-show-api-compliant?

peterahrens:
I made a rough first pass at making benchmark tools show more things. It sure shows a lot of things now. Maybe this helps with #96?


peterahrens:
sample output from my test benchmark suite
```
julia> println(IOContext(stdout, :verbose => true), results)
1-element BenchmarkTools.BenchmarkGroup:
  tags: []
  "matmatmul" => 2-element BenchmarkTools.BenchmarkGroup:
	  tags: []
	  (N = 16, M = 16, K = 16) => 3-element BenchmarkTools.BenchmarkGroup:
		  tags: []
		  "naive" => Trial(5.525 μs)
		  "default" => Trial(1.020 μs)
	  (N = 4, M = 4, K = 4) => 3-element BenchmarkTools.BenchmarkGroup:
		  tags: []
		  "naive" => Trial(179.000 ns)
		  "default" => Trial(243.000 ns)
```

jrevels:
LGTM, thanks!

Are there any places in the docs that need updating before merging this? Seems like there's at least one mention of `showcompact` in there.

peterahrens:
I think that was the only mention of show, so I updated `manual.md`.

peterahrens:
:tada:

PR #128, opened 2018-12-08
Improve/fix show methods

tkf:
Previously the logic for switching compact or verbose form of various
benchmark result objects are in `show(io::IO, group::BenchmarkGroup)`.
A more natural place would be `show(io::IO, t::Trial)` etc. as it let
code outside BenchmarkTools to switch the compactness using the
standard `IOContext` interface.  For example, with this patch `Trial`s
in `Vector` are shown in the compact form automatically.

This patch also adds:
* Colored output for the words `improvement` and `regression`.
* `:typeinfo` IO context support

As a result, you can get a nice summary view by putting benchmark objects in a `DataFrame`:

image

### DIFF
--- a/src/trials.jl
+++ b/src/trials.jl
@@ -295,7 +295,7 @@ Base.summary(io::IO, t::TrialJudgement) = withtypename(io, t) do
 end
 
 _show(io, t) =
-    if !(get(io, :verbose, false)) && get(io, :compact, true)
+    if get(io, :compact, true)
         summary(io, t)
     else
         show(io, MIME"text/plain"(), t)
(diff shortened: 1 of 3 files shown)


src/trials.jl
```diff
@@ -262,12 +273,40 @@ function prettymemory(b)
...
+Base.summary(io::IO, t::Trial) = _summary(io, t, prettytime(time(t)))
+Base.summary(io::IO, t::TrialEstimate) = _summary(io, t, prettytime(time(t)))
+Base.summary(io::IO, t::TrialRatio) = _summary(io, t, prettypercent(time(t)))
+Base.summary(io::IO, t::TrialJudgement) = withtypename(io, t) do
+    print(io, prettydiff(time(ratio(t))), " => ")
+    printtimejudge(io, t)
+end
+
+_show(io, t) =
+    if !(get(io, :verbose, false)) && get(io, :compact, true)
```
tkf: It looks like `:verbose` is added in #121 by simply copying the old `verbose` keyword argument.  Now that the standard `:compact` IO context is supported (as of #121), it seems redundant and I suggest to remove it.  What do you think?

vchuravy: Sure!

tkf: Good. It's removed now.

jrevels:
Thanks!

PR #140, opened 2019-06-21
Document the practice of referencing and dereferencing variables to stop constant prop

MasonProtter:
A common issue people run into is not knowing how to stop sufficiently simple benchmarks from being elided. 
```julia
julia> a = 1; b = 2;

julia> @btime $a + $b
  0.024 ns (0 allocations: 0 bytes)
3
```
Adding these integers at compile time can be avoided by referencing and dereferencing them:
```julia
julia> @btime $(Ref(a))[] + $(Ref(b))[]
  1.277 ns (0 allocations: 0 bytes)
3
```
but that's tricky to remember, annoying to write and hard to explain. This PR implemented a macro for doing this automatically
```julia
julia> @refd @btime $a + $b
  1.277 ns (0 allocations: 0 bytes)
3
```

I see that in the manual it was written

> In conclusion, BenchmarkTools will faithfully report the performance of the exact code that you provide to it, including any compiler optimizations that might happen to elide the code completely. It's up to you to design benchmarks which actually exercise the code you intend to exercise.

which leads me to think that this macro may not be a welcome addition to the package. I figured then that I should at least propose this PR as a work in progress before I put too much time into figuring out better semantics, documentation and adding tests so I can see if it's at least negotiable to provide the macro. 

Thoughts?

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,13 +93,6 @@ julia> @btime $(Ref(a))[] + $(Ref(b))[]
   1.277 ns (0 allocations: 0 bytes)
 3
 ```
-but this can be tricky to remember, annoying to write and hard to explain, so we provide a macro `@refd` to automatically reference and dereference interpolated variables
-```julia
-julia> @refd @btime $a + $b
-  1.277 ns (0 allocations: 0 bytes)
-3
-```
-Note that `@refd` should be appended *outside* the `@btime` statement. 
 
 As described the [manual](doc/manual.md), the BenchmarkTools package supports many other features, both for additional output and for more fine-grained control over the benchmarking process.
 
--- a/README.md
+++ b/README.md
@@ -89,7 +89,7 @@ julia> @btime $a + $b
 ```
 As a rule of thumb, if a benchmark reports that it took less than a nanosecond to perform, this hoisting probably occured. You can avoid this by referencing and dereferencing the interpolated variables 
 ```julia
-julia> @btime $(Ref(a))[] + $(Ref(a))[]
+julia> @btime $(Ref(a))[] + $(Ref(b))[]
   1.277 ns (0 allocations: 0 bytes)
 3
 ```
--- a/README.md
+++ b/README.md
@@ -78,6 +78,29 @@ julia> @btime inv(rand(3,3));     # the rand(3,3) call is included in the benchm
   1.295 μs (11 allocations: 2.47 KiB)
 ```
 
+Sometimes, interpolating variables into very simple expressions can give the compiler more information than you intended, causing it to "cheat" the benchmark by hoisting the calculation out of the benchmark code
+```julia
+julia> a = 1; b = 2
+2
+
+julia> @btime $a + $b
+  0.024 ns (0 allocations: 0 bytes)
+3
+```
+As a rule of thumb, if a benchmark reports that it took less than a nanosecond to perform, this hoisting probably occured. You can avoid this by referencing and dereferencing the interpolated variables 
+```julia
+julia> @btime Ref($a)[] + Ref($b)[]
+  1.524 ns (0 allocations: 0 bytes)
+3
+```
+but this can be annoying to remember and write, so we provide a macro `@refd` to automatically reference and dereference interpolated variables
+```julia
+julia> @refd @btime $a + $b
+  1.277 ns (0 allocations: 0 bytes)
+3
+```
+Note that `@refd` should be appended *outside* the `@btime` statement. 
+
 As described the [manual](doc/manual.md), the BenchmarkTools package supports many other features, both for additional output and for more fine-grained control over the benchmarking process.
 
 ## Why does this package exist?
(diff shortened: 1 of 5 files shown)


JeffreySarnoff:
This is helpful, I am copying it to my private resources. Adding a few tests seems appropriate.

JeffreySarnoff:
e.g.
`a=1.0; b=2.0;` 
`(@refd @belapsed $a + $b) > (@belapsed $a + $b)`

MasonProtter:
> e.g.
> `a=1.0; b=2.0;`
> `@test (@refd @belapsed $a + $b) > (@belapsed $a + $b)`

I think this sort of test is a bad idea since it can non-deterministically fail, especially on a cloud VM like travis. Instead, I added these tests:
```julia
let a=1.0, b=2.0
    @test (@macroexpand @refd $a + $b) == (@macroexpand $(Ref(a))[] + $(Ref(b))[])
    @test try (@refd @belapsed $a + $b); true catch e; false end
end
```
The first makes sure the macro is expanding correctly and the second makes sure that using `@refd` with `@belapsed` doesn't error.

JeffreySarnoff:
Coveralls wants each of these two lines covered by each of two simple tests:
https://coveralls.io/builds/24141132/source?filename=src%2Fexecution.jl#L404

with that, all checks should be successful

JeffreySarnoff:
```Error During Test at /home/travis/build/JuliaCI/BenchmarkTools.jl/test/ExecutionTests.jl:187
  Test threw exception
  Expression: walk(1, 1, (x->begin
                x + 1
            end)) == 2
  UndefVarError: walk not defined
```
it is not defined because it is an unexported thing that has not been accessed explicitly.
I think you need to use `BenchmarkTools.walk` and `BenchmarkTools._refd` in the ExecutionTests file.
Either that or `import BenchmarkTools: walk, _refd` (or both -- its been a while since I encountered this)

JeffreySarnoff:
I am in favor of this addition to `BenchmarkTools`.  There is a continuing effort by some of us to get reliable comparisons of different implementations for very fast snippets. The fact that "BenchmarkTools will faithfully report the performance of the exact code that you provide to it, including any compiler optimizations that might happen to elide the code completely. " does not preclude making available an annotation (in the spirit of `@noinline`) which simplifies this avenue of development.  I'd prefer a more obvious name to `@refd` (`@byref` for what-it-does, or  `@noelide` for how-it-operates).  To faithfully report relative performance, this provision would be of real service.

MasonProtter:
Thanks for the advice and comments Jeffrey. 

Yes, I'm certainly not committed to the name `@refd`, I just couldn't think of a better one. 

I think `@noelide`  is certainly a more descriptive name, but I worry that this might not be able to avoid eliding in all situations. I don't know enough about the compiler to say whether or not that name would be misleading. 

Perhaps `@nopropagate` or `@noconstant` since I think this thing works by disallowing constant propagation?

MasonProtter:
One other thing worth pondering in the design of this: Currently the macro references and dereferences only interpolated variables. 

We may want to make it work on non-interpolated variables or literals as well. I'm still undecided on this.

JeffreySarnoff:
Give me an example where it is useful to `enreference` a literal or a variable that has not been `$` prefixed.

MasonProtter:
Someone showed up on slack just half an hour ago asking why `log` was so much slower than `exp` and showed this code example
```
julia> @btime log(10.0)
  3.139 ns (0 allocations: 0 bytes)
2.302585092994046

julia> @btime exp(10.0)
  0.023 ns (0 allocations: 0 bytes)
22026.465794806718
```
For this code, (de)referencing `10` stops the hoisting:
```
julia> @btime exp(Ref(10)[])
  7.116 ns (0 allocations: 0 bytes)
22026.465794806718
```
though of course it's better to initialize a variable and then interpolate it in and then (de)reference that.

JeffreySarnoff:
yes. while your PR does ring compatible with the intent of and consonant with the direction of benchmark tools -- this feels unbenchmark-tooling-like.  Its not about saving them from their own inexperience; this is about furthering the productivity of those with experience by simplifying access to properly non-elideded performance metrics without putting anything in the way benchmark tools now operates and is used.

MasonProtter:
Fair enough, I just figured I should at least raise the question of what sort of symbols the macro should be operating on.

JeffreySarnoff:
There is something wrong the inline docs that you provide for "refd".  Have you tried calling up that help.  One of the best sources to copy for that is in the source files [here](https://github.com/JuliaDocs/Documenter.jl/blob/master/src/)

MasonProtter:
Are you referring to 
```
help?> @refd
  @refd @btime expression

  or @refd @belapsed expression or @refd @benchmark expression
```
 or another problem? I had missed this one until now. Will fix tomorrow. Do you have any suggestions for how to phrase that? I feel like writing
```
    @refd @btime expression

  or 

    @refd @belapsed expression 

or 

    @refd @benchmark expression
```
is a bit awkward.

JeffreySarnoff:
yes that 
"""
     @referred _bmacro_ expression

where  _bmacro_ is  @benchmark, @btime, or  @belapsed
"""

(dunno why this messes up the formatting -- ignore that)

vchuravy:
I am unsure about adding the `macro`, but I think we certainly should have the documentation addition.

@jrevels do you have a strong opinion?

vchuravy:
Thanks Mason!

MasonProtter:
Sorry I kind forgot about it, but better late than never I guess :)

PR #142, opened 2019-07-30
Add more granular functions for time and memory judgements

DilumAluthge:
## Summary

Currently, BenchmarkTools defines the following functions:
- `isinvariant`
- `isregression`
- `isimprovement`
- `invariants`
- `regressions`
- `improvements`

This pull request adds the following new functions:
- `istimeinvariant`
- `ismemoryinvariant`
- `istimeregression`
- `ismemoryregression`
- `istimeimprovement`
- `ismemoryimprovement`
- `timeinvariants`
- `memoryinvariants`
- `timeregressions`
- `memoryregressions`
- `timeimprovements`
- `memoryimprovements`

Additionally, it redefines `isinvariant` in terms of `istimeinvariant` and `ismemoryinvariant`, `isregression` in terms of `istimeregression` and `ismemoryregression`, and `isimprovement` in terms of `istimeimprovement` and `ismemoryimprovement`.

## Motivation

The existing function `isregression` is convenient if you want to know about any kind of regression. However, there may be situations in which you only care about time regressions or only care about memory regressions. In those cases, the `isregression` function is not helpful, and more granular functions such as `istimeregression` and `ismemoryregression` will be needed. The same logic goes for the other new functions.


DilumAluthge:
@vchuravy @ararslan Can you review this?

DilumAluthge:
Bump.

If there are no objections, @ararslan or @vchuravy could you merge this?

KristofferC:
These long  single word names are quite hard to read.

ararslan:
That's a fair point. I also wonder if perhaps `isregression` could return a named tuple, say `(time=true, memory=false)`. (Just spitballing.)

DilumAluthge:
> That's a fair point. I also wonder if perhaps isregression could return a named tuple, say (time=true, memory=false). (Just spitballing.)

Personally, when I see a function named `isBLANK`, I assume that it's going to return a Bool. Also, if we change the return type of `isregression`, it'll be a breaking change. I was hoping to avoid having to make a breaking change of `BenchmarkTools`.

> Arguably some keyword to the original functions might be cleaner.

I think this could work. We'll have keyword arguments `time::Bool` and `memory::Bool`, with default values `time = true` and `memory = true`. I'll make those changes now.

ararslan:
> with default values `time = true` and `memory = true`

That means you call it like `isregression(..., memory=false)`, which reads weird.

DilumAluthge:
> That means you call it like isregression(..., memory=false), which reads weird.

Yeah that is very weird.

DilumAluthge:
We could just add underscores for the long names, and leave the short names like `isregression` unchanged?

DilumAluthge:
Alternatively, we could introduce immutable singleton types `Time` and `Memory`.

Then `isregression(Time, ...)` would only look at time, and `isregression(Memory, ...)` would only look at memory. `isregression(...)` would preserve the current behavior.

DilumAluthge:
The other option is to add underscores to all of the names. But that would require us to make the next release of `BenchmarkTools` a breaking release.

ararslan:
I don't personally find the names introduced here to be that bad.

DilumAluthge:
It's up to you guys! I'm happy to make whatever changes. I would like this functionality.

DilumAluthge:
Bump. Would it be possible to merge this?

The reason I ask is that my pull request https://github.com/JuliaCI/PkgBenchmark.jl/pull/85 depends on the changes in this pull request.

DilumAluthge:
Here's another option for the keyword arguments:

```julia
isregression(group::BenchmarkGroup; type = :time) 
isregression(group::BenchmarkGroup; type = :memory)
isregression(group::BenchmarkGroup; type = :both)
isregression(group::BenchmarkGroup) # equivalent to `type = :both`
```

And similarly for the other functions.

DilumAluthge:
@vchuravy I like that idea. I have made those changes. Take a look and let me know what you think.

ararslan:
> Could you please squash the commits

Alternatively we can use GitHub's squash-merge

vchuravy:
I am a merge-commit fanatic, but sure

DilumAluthge:
Thanks for merging guys! Sorry I didn't see this until now.

PR #147, opened 2019-09-23
add @benchmarkset & @case for defining benchmark suite

Roger-luo:
This PR adds 2 macros to provide a simpler syntax for defining large benchmark suite.

I found it can be a bit annoying if there are multiple benchmark suite like this: 

https://github.com/QuantumBFS/YaoArrayRegister.jl/blob/master/benchmark/benchmarks.jl

it is not necessary to manually control things like `SUITE["matrices"]["contiguous"]`, thus I implement these two macro to make defining benchmark suite in a similar syntax as the `testset` in `Test`.

now you can just write things like

```jl
@suite "suite 1" begin
     @case k for k in 1:n
            rand($k,  $k) * rand($k, $k)
      end
end
```

instead of

```jl
SUITE = BenchmarkGroup()
SUITE["suite 1"] = BenchmarkGroup()

for k in 1:n
    SUITE["suite 1"][k] = @benchmarkable rand($k, $k) * rand($k, $k)
end
```

PS. currently it will generate a constant `SUITE`, which is compatible with `PkgBenchmark`

Update: I find I need some more modification on this new syntax, the current one still fails to deal with some cases.


Roger-luo:
I changed the name to `@benchmarkset` and `@case`, this will provide a similar way to define benchmark suite to `@testset` and `@test`

vchuravy:
Can you rebase onto master to get CI?

vchuravy (approved):

Roger-luo:
bump just fixed the tests.

Roger-luo:
hi @vchuravy mind releasing a patch version for this?

KristofferC:
Just pointing out, this package was intentionally designed to not use macros since they were considered confusing and unclear:

https://github.com/JuliaCI/PkgBenchmark.jl/issues/9#issuecomment-257952067

Roger-luo:
> Just pointing out, this package was intentionally designed to not use macros since they were considered confusing and unclear:

@KristofferC  Ah that makes sense, I didn't know one can use this package with `addgroup!` , it seems not being mentioned in manual at all, only in https://github.com/JuliaCI/BenchmarkTools.jl/blob/master/doc/reference.md#misc-functions

I feel need to polish the documentation a bit for this if we want to revert this PR. I'm OK with both, not sure what other people thinks.

KristofferC:
Well, too late now.

awadell1:
Is the equivalent using `addgroup!` something like:

```julia

# Root Suite
SUITE = BenchmarkGroup()

# BenchmarkGroup generated by @benchmarkset
s = BenchmarkGroup()

# For loop for each @case 
for k in 1:n
    s[k] = @benchmarkable rand($k, $k) * rand($k, $k)
end

addgroup!(SUITE, s)
```

PR #149, opened 2019-09-30
reexport mean and median

milesfrain:


src/BenchmarkTools.jl
```diff
@@ -2,11 +2,8 @@ module BenchmarkTools
 
 using JSON
 using Base.Iterators
-
-using Statistics
```
ararslan: We still want to be `using Statistics` here and extending the functions from Statistics; we can simply add `mean` and `median` to the exports, which will reexport the right things.

milesfrain: Good suggestion. Updated latest PR.

src/BenchmarkTools.jl
```diff
@@ -46,6 +48,8 @@ export BenchmarkGroup,
        invariants,
        regressions,
        improvements,
+       median,
```
ararslan: Listed twice?

milesfrain: I assumed a separate export was need to for both `::Trial` and `::BenchmarkGroup`, but with the magic of Julia, it seems like only one is necessary. Will update to list these only under Trial.

Issue #118, opened 2018-09-21
Simple version of @btime with automatic interpolation


jlperla:
I know this has been brought up before (e.g. https://github.com/JuliaCI/BenchmarkTools.jl/issues/65 ) and others, but let me make a slightly different suggestion.  The current `@btime` and `@benchmark` are great, but what about a simple @rdeits made a similar suggestion in https://github.com/JuliaCI/BenchmarkTools.jl/issues/65#issuecomment-313937076 

The issue is that the vast majority of benchmarking I do is of the form `@btime f(a,b)` and no matter how many times I have read the manual, I will always forget to do it as `@btime f($a, $b)`.  Not to mention that it is harder to read or flip benchmarks on/off easily for a function you are working on.

So my suggestion is a macro `@bfunction` or whatever, which only accepts a function as head and automatically interpolates all arguments.  i.e. 
``
`@bfunction f(a,b) -> @btime f($a, $b)
```
Is that sort of thing possible to do with the macro expansion model in Julia?  If I was able to get a PR would you consider merging it, or do you disagree with the feature?

jlperla:
Why a new package rather than a new macro within this package? Is there resistance to merge something of that sort, or something more.

Regardless, if it is in a new package (and your name sounds fine) then I don't think it should clashes of named macros with this package, just in case something uses both, or gets confused on behaviour

jlperla:
> Honestly, I think the root of the problem is just that scope is a very important part of benchmarking in Julia, and one cannot author a well-defined benchmark (or correctly interpret results) without explicitly understanding the scope of data manipulated by the benchmark. 

I think you are exactly right.  And that is plenty of reason that it shouldn't replace `@btime`, but be something distinct (and likely limited in scope/features).

The issue is that `@btime` is being used for things other than ``well-defined benchmarks`` and instead as an exploratory tool and rough check on performance.  For most users, this is what they are looking for, and they are willing to sacrifice all sorts of features as long as it gets it right most of the time.  This exploratory use-case also tends to be one which is toggled on/off on a function, and the manual changing of the `$` is easy to forget.

Right now the heuristic people like myself are using is "always put `$` in front of everything... unless I forget".

rdeits:
Yeah, I noticed the issue with `setup` right after I posted that. Fortunately, it's not too bad to work around: rather than interpolating everything, I just add entries to `setup` for any local symbols (except those symbols which already have entries in `setup`). So if we have: 

```julia
[...]

PR #150, opened 2019-10-01
WIP: rebasing @localbtime and friends from LocalScopeBenchmarks.jl

miguelraz:
This is a WIP PR to merge a `@localbtime` and friends macro into BenchmarkTools.jl .

There's a thread with [reported issues here](https://github.com/JuliaCI/BenchmarkTools.jl/issues/118#issuecomment-526712519) and this PR took existing code from `LocalScopeBenchmarks.jl`, with all credit to @rdeits.

This PR:
- adds a dependency on MacroTools and Ordered collections for the local sourcewalking helper functions
- exports `@localbenchmark`, `@localbtime`, `@localbelapsed`
- adds tests for said functions.
- bumps the end position semver number by `1`
- changes the statistical metric for the local benchmarks to be the mean and not the median (seemed  more robust locally, modulo CI woes.)

All feedback is welcome.

Another welcome feedback will be on how the new tests for local time regression measurement should be done: I found that when running them locally, the first run would pass all tests, then fail due only to the regression threshold set. I think it's hitting the microbenchmarks remembering cache history.

Lastly, this PR does not:
- make a judgement on how to/if at all to deprecate `@btime` et al. 

That is left as a discussion for another time, and for now I will focus on landing this functionality.

Ping also to @jlperla, @chrisrackauckas.

### DIFF
--- a/test/LocalScopeBenchmarkTests.jl
+++ b/test/LocalScopeBenchmarkTests.jl
@@ -30,11 +30,13 @@ global_x = 1.0
         @test isinvariant(j)
     end
 
-        @testset "Generated code is identical" begin
-            x = 1.0
-            ex1 = Meta.@lower(@benchmark($sin($x), evals=500))
-            ex2 = Meta.@lower(@localbenchmark(sin(x), evals=500))
-        end
+# This test fails to run if copy/pasted into the REPL due to differing LineNumbers where vars get
+# pulled from.
+    @testset "Generated code is identical" begin
+        x = 1.0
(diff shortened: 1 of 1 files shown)


miguelraz:
The tests are failing / and passing due to a brittle judging function. One idea would be to judge loosely min, mean and median and if 2/3 pass then the judgement is considered invariant.

gdalle:
Hey @miguelraz, just got ~~onboarded~~ trapped as a maintainer, and I wanted to see if this was still relevant some 4 years later?

PR #157, opened 2020-02-16
Add ballocated macro

oxinabox:
Matchs to the `@allocated` macro.
Which I used in a few tests,
but which in Julia 1.4+ now disagrees with allocations reported by `@btime`
(see https://github.com/invenia/NamedDims.jl/issues/89 caused by https://github.com/JuliaLang/julia/pull/33717)

So I want `@ballocated`: allocation sizes that I can trust.

### DIFF
--- a/src/execution.jl
+++ b/src/execution.jl
@@ -375,7 +375,7 @@ this returns the number of bytes allocated when executing
 a given expression.   It uses the `@benchmark`
 macro, however, and accepts all of the same additional
 parameters as `@benchmark`.  The returned allocations
-is for the trial with the *minimum* elapsed time measured
+correspond to the trial with the *minimum* elapsed time measured
 during the benchmark.
 """
 macro ballocated(args...)
--- a/src/execution.jl
+++ b/src/execution.jl
@@ -348,7 +348,7 @@ end
 ######################
 
 # These macros provide drop-in replacements for the
-# Base.@time and Base.@elapsed macros, which use
+# Base.@time, Base.@elapsed macros and Base.@allocated, which use
 # @benchmark but yield only the minimum time.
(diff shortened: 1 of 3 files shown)


oxinabox:
I would like this, as I am currently using it in the tests for [AutoPreallocation.jl](https://github.com/oxinabox/AutoPreallocation.jl)

src/execution.jl
```diff
@@ -367,6 +367,23 @@ macro belapsed(args...)
...
 
+"""
+    @ballocated expression [other parameters...]
+
+Similar to the `@allocated` macro included with Julia,
+this returns the number of bytes allocated when executing
+a given expression.   It uses the `@benchmark`
+macro, however, and accepts all of the same additional
+parameters as `@benchmark`.  The returned allocations
+is for the trial with the *minimum* elapsed time measured
```
ararslan:
```suggestion
correspond to the trial with the *minimum* elapsed time measured
```

src/execution.jl
```diff
@@ -367,6 +367,23 @@ macro belapsed(args...)
...
+during the benchmark.
+"""
+macro ballocated(args...)
+    return esc(quote
+        $BenchmarkTools.memory($BenchmarkTools.minimum($BenchmarkTools.@benchmark $(args...)))
```
ararslan: Hmm, I wonder whether minimum should be configurable instead?

oxinabox: Maybe maybe not but either way
out of scope for this PR.

That would need to change brime and belapsed also And I would think it's more interesting for them

ararslan: Ah if that's what `@btime` does too then that's fine.

PR #187, opened 2021-01-08
Fix links

rmsrosa:
Fix links to "`@time` macro" and to "avoid the problems of benchmarking with globals"


rmsrosa:
Sorry, I didn't find any issue about that, but I forgot to look at the open pull-requests. I just found out there is already one with these fixes, so I will close this pull-request.

vchuravy:
No worries.

PR #189, opened 2021-01-10
switch to GitHub actions

Roger-luo:
Travis is down, I guess it's better to switch to github actions.

edit: I'm not sure if we want to set up Documenter (perhaps we should?), I commented out related parts for that, but if anyone with access wants to setup Documenter please feel free to edit this PR.


Roger-luo:
umm I'm not sure how to fix the other test on 1.6+... but perhaps someone could merge this PR first and fix it in other PRs? since it's not very related

vchuravy:
thanks some CI is better than no CI :)

PR #193, opened 2021-02-26
Early-convert BenchmarkGroup names to JSON-safe

vtjnash:
It turns out we're missing a large chunk of tuning, since the JSON file forces all keys to be string, but we expect keys to be semi-rich types (for rich filtering). Compromise instead and coerce some types to string immediately and parse-eval other strings back to basic types (string, number, and tuples thereof).

### DIFF
--- a/src/groups.jl
+++ b/src/groups.jl
@@ -3,12 +3,13 @@
 ##################
 
 struct BenchmarkGroup
-    tags::Vector{Any}
-    data::Dict{Any,Any}
+    tags::Vector{String}
+    data::Dict{String,Any}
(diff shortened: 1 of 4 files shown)


vtjnash:
The de-JSON recover-y should also incidentally fix the awkward doubled-quoting of the reports in Nanosoldier such as https://github.com/JuliaCI/NanosoldierReports/blob/master/benchmark/by_date/2021-02/17/report.md

ararslan (approved):
Makes sense I think, though it's been ages since I've thought about the internals here.

PR #194, opened 2021-02-26
Add plot capability

gustaphe:
Adds a recipe to visualize the timings of a `Trial` or a `BenchmarkGroup` of `Trial`s using boxplots. Only pulls in `RecipesBase.jl` so should be fairly low impact (but I don't know how best to check that).

```julia
using BenchmarkTools, Plots, StatsPlots
...
result = run(benchmarkgroup)
plot(result, outliers=false)
```

benchmarks

### DIFF
--- a/BenchmarkPlots/Project.toml
+++ b/BenchmarkPlots/Project.toml
@@ -1,0 +1,11 @@
+name = "BenchmarkPlots"
+uuid = "ab8c0f59-4072-4e0d-8f91-a91e1495eb26"
+version = "0.1.0"
+
+[deps]
+BenchmarkTools = "6e4b80f9-dd63-53aa-95a3-0cdb28fa8baf"
+RecipesBase = "3cdcf5f2-1ef4-517c-9805-6587b60abb01"
+
+[compat]
+BenchmarkTools = "<1"
+RecipesBase = "1"
--- a/BenchmarkPlots/Project.toml
+++ b/BenchmarkPlots/Project.toml
@@ -7,5 +7,5 @@ BenchmarkTools = "6e4b80f9-dd63-53aa-95a3-0cdb28fa8baf"
 RecipesBase = "3cdcf5f2-1ef4-517c-9805-6587b60abb01"
 
 [compat]
-BenchmarkTools = "<1"
+BenchmarkTools = "1"
 RecipesBase = "1"
--- a/BenchmarkPlots/src/BenchmarkPlots.jl
+++ b/BenchmarkPlots/src/BenchmarkPlots.jl
@@ -1,0 +1,25 @@
+module BenchmarkPlots
+using RecipesBase
+using BenchmarkTools: Trial, BenchmarkGroup
+
+@recipe function f(::Type{Trial}, t::Trial)
+    seriestype --> :violin
+    legend --> false
+    yguide --> "t / ns"
+    xticks --> false
+    t.times
+end
+
+@recipe function f(g::BenchmarkGroup, keys=keys(g))
+    legend --> false
+    yguide --> "t / ns"
+    for k in keys
+        @series begin
+            label --> string(k)
+            xticks --> true
+            [string(k)], g[k]
+        end
+    end
+end
+
+end
--- a/Project.toml
+++ b/Project.toml
@@ -11,7 +11,7 @@ UUIDs = "cf7118a7-6976-5b1a-9a39-7adc72f591a4"
 
 [compat]
 julia = "1"
-JSON  = "0.18, 0.19, 0.20, 0.21"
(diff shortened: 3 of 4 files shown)


vchuravy:
Thank you for your PR! I am hesitant to take on any extra dependencies for BenchmarkTools, right now we only depend on JSON.jl 

But I agree with you that better visualization would be really nice. I have hacked up to many ad-hoc plots before :) Maybe a companion package?

Lastly, I am not sure if a boxplot is the best visualization strategy since benchmark results are often not uni-modal. So if you get a multi-modal distribution that wouldn't show up in a boxplot.

gustaphe:
I don't think I know of any universally accepted multi-modal visualization, but you can 
```julia
plot(result; st=:violin)
```
and it'll do the right thing. Maybe that should be the default, I haven't run so advanced benchmarks that I've needed more than a quick glance.

Would anyone go through the trouble of finding and loading `BenchmarkToolPlots`? It feels like the kind of functionality that you check if it just works, and when you find it doesn't you hack something together. I could be wrong.

vchuravy:
Ah violin plot might work quite nicely.

We could emphasize the visualization packages in the readme and have the examples use it! 
(It could also live in this repo, but as a sub-package). Keeping the dependency list so small is an intentional feature for BenchmarkTools since it is generally at the bottom of the dependency chain.

gustaphe:
Sure, I can draft up a pr like that one of these days.

gustaphe:
This is difficult to test, I haven't done any subdir-package stuff before, but there's a process:

https://github.com/JuliaRegistries/Registrator.jl#registering-a-package-in-a-subdirectory

gustaphe:
Now one can choose the order of the trials:

```julia
using BenchmarkTools, BenchmarkPlots, StatsPlots
group = BenchmarkGroup()
group[:apple] = @benchmarkable sort!(a) setup=(a=rand(1000))
group[:banana] = @benchmarkable sort!(a) setup=(a=rand(1100))
group[:pineapple] = @benchmarkable sort!(a) setup=(a=rand(900))
trial = run(group)
plot(
    plot(trial),
    plot(trial, [:pineapple, :apple, :banana]),
    )
```

benchmarkplots

vchuravy:
Thanks! This looks great :) Would you mind rebasing?

gustaphe:
How's this? I guess no bump in version, because the base package only has docs changes.

BenchmarkPlots/Project.toml
```diff
@@ -0,0 +1,11 @@
+name = "BenchmarkPlots"
+uuid = "ab8c0f59-4072-4e0d-8f91-a91e1495eb26"
+version = "0.1.0"
+
+[deps]
+BenchmarkTools = "6e4b80f9-dd63-53aa-95a3-0cdb28fa8baf"
+RecipesBase = "3cdcf5f2-1ef4-517c-9805-6587b60abb01"
+
+[compat]
+BenchmarkTools = "<1"
```
vchuravy, line 10:
```suggestion
BenchmarkTools = "1"
```

vchuravy:
We should figure out a way to setup tests.

mcabbott:
Should this sub-package depend on StatsPlots? 
```julia
julia> b = @benchmark copy(rand(128));

julia> using BenchmarkPlots

julia> plot(b)
┌ Warning: seriestype violin has been moved to StatsPlots.  To use: `Pkg.add("StatsPlots"); using StatsPlots`
└ @ Plots ~/.julia/packages/Plots/AJMX6/src/args.jl:1575
ERROR: The backend must not support the series type Val{:violin}, and there isn't a series recipe defined.

(jl_Nt1QFi) pkg> st
Status `/private/var/folders/yq/4p2zwd614y59gszh7y9ypyhh0000gn/T/jl_Nt1QFi/Project.toml`
  [ab8c0f59] BenchmarkPlots v0.1.0
  [91a5bcdd] Plots v1.25.0
```

vchuravy:
Probably...

gustaphe:
In theory, you can `plot(b; st=:hist)`, or some other series type not from statsplots. But yes, in practice that would probably be simpler. 

The next question is, does `using StatsPlots` from `BenchmarkPlots` make the seriestypes available in the outer scope?

mcabbott:
I presume that would work, as Plots uses Revise which would notice, but I'm not sure without trying.

I guess the other option would be to make the default be a histogram, since that works.

However, I think the histogram has mis-labeled axes? Also, I need to set limits by hand to see anything, in the common case that the maximum is huge. And when I do so, the violin plot seems to have been down-sampled to not show very much?

Screenshot 2021-12-05 at 22 17 08

PR #195, opened 2021-03-04
record exact Julia commit in JSON

vtjnash:
For added clarity, and https://github.com/JuliaCI/Nanosoldier.jl/issues/23


ararslan (approved):
Seems harmless enough :shrug:

vtjnash:
I decided I don't think this is the right place for it. Probably we should record something like Base.pkgorigins here instead, if anything

PR #209, opened 2021-05-19
set up Documenter

Roger-luo:
this package deserves a documenter page... not sure if we should trim README.md a bit now since quick start goes into the index.md now


DilumAluthge:
I've added a `DOCUMENTER_KEY`.

PR #212, opened 2021-05-21
Example for storing calculated value to a variable

StatisticalMice:
I created this PR based on the discussion in Slack. I believe this would be useful for beginners.

### DIFF
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -38,6 +38,17 @@ julia> @btime sin(x) setup=(x=rand())
 0.49587200950472454
 ```
 
+If you want to benchmark an expression and store the value in a variable, you can use:
+
+```julia
+julia> y = @btime sin(x) setup=(x=rand())
+  3.467 ns (0 allocations: 0 bytes)
+0.3812650449427549
+
+julia> y
+0.3812650449427549
+```
+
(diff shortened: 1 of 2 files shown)


docs/src/index.md
```diff
@@ -38,6 +38,17 @@ julia> @btime sin(x) setup=(x=rand())
...
+
+```julia
+julia> y = @btime sin(x) setup=(x=rand())
+  3.467 ns (0 allocations: 0 bytes)
+0.3812650449427549
+
+julia> y
+0.3812650449427549
+```
+
```
giordano, line 51: `@btime` docstring already documents that it returns the value, as `@time` also does, but I guess adding an explicit example doesn't hurt (which could also be added to `@btime` docstring directly instead).  This specific example, however, doesn't look extremely useful to me: you're taking the sine of a random value, well, one of the many several random values being generated for the benchmark.  Is that likely to be a realistic situation?

gdalle:
Closing because the current docs mentions this: 

> Alternatively, you can use the @btime or @belapsed macros. These take exactly the same arguments as @benchmark, but behave like the @time or @elapsed macros included with Julia: @btime prints the minimum time and memory allocation before returning the value of the expression, while @belapsed returns the minimum time in seconds.

Issue #146, opened 2019-09-20
Change median and mean export, or update documentation


milesfrain:
There's a mismatch between what the documentation suggests and how these functions are exported.

`maximum` and `minimum` are exported to `Base`, while `median` and `mean` are exported to `Statistics`.
https://github.com/JuliaCI/BenchmarkTools.jl/blob/65ef038de1c1ffa0f5d4f4088044214457f0f7df/src/groups.jl#L66-L69

The [manual](https://github.com/JuliaCI/BenchmarkTools.jl/blob/master/doc/manual.md#trial-and-trialestimate) does not mention the need for `using Statistics` to get the the remaining two functions to work.

The [reference guide](https://github.com/JuliaCI/BenchmarkTools.jl/blob/master/doc/reference.md#medianxuniontrial-benchmarkgroup) also suggests that `median` and `mean` are exported the same as `minimum` and `maximum`.

Unsure whether `median` and `mean` should be exported to `Base`, since this wouldn't extending an existing `Base` method.

Similar to #143

ararslan:
`median` and `mean` were _removed_ from Base and put into Statistics. We should either reexport those or tell people to run `using Statistics` in the documentation.

CameronBieganek:
I just bumped into this again. I had already forgotten that I needed to do `using Statistics` to get `median` and `mean` to work, and the **BenchmarkTools** documentation didn't remind me...

ararslan:
A PR to document or reexport `mean` and `median` would be welcome. I'm fairly indifferent on which is preferable, though I guess I have a slight inclination toward the latter.

milesfrain:
See #149 for the latter option

milesfrain:
Is there additional documentation required? I thought the fix to re-export now ensures that the existing docs are now aligned with the code.

gdalle:
Maybe just a remark stating that they're re exported is what I had in mind

milesfrain:
Add the remark and applied some additional edits in #336

PR #217, opened 2021-05-28
Overhaul display of a Trial

### LINKED ISSUE #215: A more compact, colourful, output style.
Hi!

I find this tool great to use, but I can't help but think that the visual output could be polished a bit. One micro-benchmark tool that I think does a really nice job with this is [hyperfine](https://github.com/sharkdp/hyperfine).

image

I think looking at what's done well by it could lead to a Benchmark output that's a bit easier to glance at :slightly_smiling_face: 

For example:

image

which I think is nicer than

image

tecosaur:
Use layout inspired by hyperfine, with an added histogram. Closes #215.

Sample:

<details>
<summary>
Initial version
</summary>

image
</details>

image

### DIFF
--- a/docs/src/manual.md
+++ b/docs/src/manual.md
@@ -25,17 +25,16 @@ To quickly benchmark a Julia expression, use `@benchmark`:
 
 ```julia
 julia> @benchmark sin(1)
-BenchmarkTools.Trial:
-  memory estimate:  0 bytes
-  allocs estimate:  0
-  --------------
-  minimum time:     13.610 ns (0.00% GC)
-  median time:      13.622 ns (0.00% GC)
-  mean time:        13.638 ns (0.00% GC)
-  maximum time:     21.084 ns (0.00% GC)
-  --------------
-  samples:          10000
-  evals/sample:     998
+BechmarkTools.Trial: 10000 samples with 1000 evaluations.
+ Range (min … max):  1.442 ns … 53.028 ns  ┊ GC (min … max): 0.00% … 0.00%
+ Time  (median):     1.453 ns              ┊ GC (median):    0.00%
+ Time  (mean ± σ):   1.462 ns ±  0.566 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%
+
+                                   █                              
+  ▂▁▁▃▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁█▁▁█▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▃▁▁▃
+  1.44 ns           Histogram: frequency by time           1.46 ns (top 1%)
+
+ Memory estimate: 0 bytes, allocs estimate: 0.
 ```
 
 The `@benchmark` macro is essentially shorthand for defining a benchmark, auto-tuning the benchmark's configuration parameters, and running the benchmark. These three steps can be done explicitly using `@benchmarkable`, `tune!` and `run`:
@@ -47,17 +46,16 @@ julia> b = @benchmarkable sin(1); # define the benchmark with default parameters
 julia> tune!(b);
 
 julia> run(b)
-BenchmarkTools.Trial:
-  memory estimate:  0 bytes
-  allocs estimate:  0
-  --------------
-  minimum time:     13.605 ns (0.00% GC)
-  median time:      13.618 ns (0.00% GC)
-  mean time:        13.631 ns (0.00% GC)
-  maximum time:     22.383 ns (0.00% GC)
-  --------------
-  samples:          10000
-  evals/sample:     998
+BechmarkTools.Trial: 10000 samples with 1000 evaluations.
+ Range (min … max):  1.442 ns … 4.308 ns  ┊ GC (min … max): 0.00% … 0.00%
+ Time  (median):     1.453 ns             ┊ GC (median):    0.00%
+ Time  (mean ± σ):   1.456 ns ± 0.056 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%
+
+                                  █                              
+  ▂▁▃▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁█▁▁█▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▂▁▁▃
+  1.44 ns          Histogram: frequency by time           1.46 ns (top 1%)
+
+ Memory estimate: 0 bytes, allocs estimate: 0.
 ```
 
 Alternatively, you can use the `@btime` or `@belapsed` macros.
@@ -112,32 +110,30 @@ You can interpolate values into `@benchmark` and `@benchmarkable` expressions:
 ```julia
 # rand(1000) is executed for each evaluation
 julia> @benchmark sum(rand(1000))
-BenchmarkTools.Trial:
-  memory estimate:  7.94 KiB
-  allocs estimate:  1
-  --------------
-  minimum time:     1.566 μs (0.00% GC)
-  median time:      2.135 μs (0.00% GC)
-  mean time:        3.071 μs (25.06% GC)
-  maximum time:     296.818 μs (95.91% GC)
-  --------------
-  samples:          10000
-  evals/sample:     10
+BechmarkTools.Trial: 10000 samples with 10 evaluations.
+ Range (min … max):  1.153 μs … 142.253 μs  ┊ GC (min … max): 0.00% … 96.43%
+ Time  (median):     1.363 μs               ┊ GC (median):    0.00%
+ Time  (mean ± σ):   1.786 μs ±   4.612 μs  ┊ GC (mean ± σ):  9.58% ±  3.70%
+
+   ▄▆██▇▇▆▄▃▂▁                           ▁▁▂▂▂▂▂▂▂▁▂▁              
+  ████████████████▆▆▇▅▆▇▆▆▆▇▆▇▆▆▅▄▄▄▅▃▄▇██████████████▇▇▇▇▆▆▇▆▆▅▅▅▅
+  1.15 μs         Histogram: log(frequency) by time          3.8 μs (top 1%)
+
+ Memory estimate: 7.94 KiB, allocs estimate: 1.
 
 # rand(1000) is evaluated at definition time, and the resulting
 # value is interpolated into the benchmark expression
 julia> @benchmark sum($(rand(1000)))
-BenchmarkTools.Trial:
-  memory estimate:  0 bytes
-  allocs estimate:  0
-  --------------
-  minimum time:     101.627 ns (0.00% GC)
-  median time:      101.909 ns (0.00% GC)
-  mean time:        103.834 ns (0.00% GC)
-  maximum time:     276.033 ns (0.00% GC)
-  --------------
-  samples:          10000
-  evals/sample:     935
+BechmarkTools.Trial: 10000 samples with 963 evaluations.
+ Range (min … max):  84.477 ns … 241.602 ns  ┊ GC (min … max): 0.00% … 0.00%
+ Time  (median):     84.497 ns               ┊ GC (median):    0.00%
+ Time  (mean ± σ):   85.125 ns ±   5.262 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%
+
+  █                                                                 
+  █▅▇▅▄███▇▇▆▆▆▄▄▅▅▄▄▅▄▄▅▄▄▄▄▁▃▄▁▁▃▃▃▄▃▁▃▁▁▁▁▁▃▁▁▁▁▁▁▁▁▁▁▃▃▁▁▁▃▁▁▁▁▆
+  84.5 ns         Histogram: log(frequency) by time           109 ns (top 1%)
+
+ Memory estimate: 0 bytes, allocs estimate: 0.
 ```
 
 A good rule of thumb is that **external variables should be explicitly interpolated into the benchmark expression**:
@@ -147,31 +143,29 @@ julia> A = rand(1000);
 
 # BAD: A is a global variable in the benchmarking context
 julia> @benchmark [i*i for i in A]
-BenchmarkTools.Trial:
-  memory estimate:  7.95 KiB
-  allocs estimate:  2
-  --------------
-  minimum time:     13.154 μs (0.00% GC)
-  median time:      13.806 μs (0.00% GC)
-  mean time:        14.071 μs (0.00% GC)
-  maximum time:     337.462 μs (0.00% GC)
-  --------------
-  samples:          10000
-  evals/sample:     1
+BechmarkTools.Trial: 10000 samples with 54 evaluations.
+ Range (min … max):  889.241 ns … 29.584 μs  ┊ GC (min … max):  0.00% … 93.33%
+ Time  (median):       1.073 μs              ┊ GC (median):     0.00%
+ Time  (mean ± σ):     1.296 μs ±  2.004 μs  ┊ GC (mean ± σ):  14.31% ±  8.76%
+
+      ▃█▆                                                           
+  ▂▂▄▆███▇▄▄▃▃▃▃▃▂▂▂▂▂▂▂▂▂▂▂▁▂▂▂▁▂▂▁▁▁▁▁▂▁▁▁▁▂▂▁▁▁▁▂▁▁▁▁▁▁▂▂▂▂▂▂▂▂▂▂
+  889 ns             Histogram: frequency by time            2.92 μs (top 1%)
+
+ Memory estimate: 7.95 KiB, allocs estimate: 2.
 
 # GOOD: A is a constant value in the benchmarking context
 julia> @benchmark [i*i for i in $A]
-BenchmarkTools.Trial:
-  memory estimate:  7.95 KiB
-  allocs estimate:  2
-  --------------
-  minimum time:     929.375 ns (0.00% GC)
-  median time:      1.348 μs (0.00% GC)
-  mean time:        2.405 μs (36.64% GC)
-  maximum time:     91.481 μs (95.46% GC)
-  --------------
-  samples:          10000
-  evals/sample:     32
+BechmarkTools.Trial: 10000 samples with 121 evaluations.
+ Range (min … max):  742.455 ns … 11.846 μs  ┊ GC (min … max):  0.00% … 88.05%
+ Time  (median):     909.959 ns              ┊ GC (median):     0.00%
+ Time  (mean ± σ):     1.135 μs ±  1.366 μs  ┊ GC (mean ± σ):  16.94% ± 12.58%
+
+  ▇█▅▂                                                             ▁
+  ████▇▃▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▄▅▆██
+  742 ns          Histogram: log(frequency) by time          10.3 μs (top 1%)
+
+ Memory estimate: 7.94 KiB, allocs estimate: 1.
 ```
 
 (Note that "KiB" is the SI prefix for a [kibibyte](https://en.wikipedia.org/wiki/Kibibyte): 1024 bytes.)
@@ -187,18 +181,18 @@ julia> b = @benchmarkable fill!($A, rand());
 julia> run(b, samples = 1);
 
 julia> A
-3-element Array{Float64,1}:
- 0.837789
- 0.837789
- 0.837789
+3-element Vector{Float64}:
+ 0.4615582142515109
+ 0.4615582142515109
+ 0.4615582142515109
 
 julia> run(b, samples = 1);
 
 julia> A
-3-element Array{Float64,1}:
- 0.647885
- 0.647885
- 0.647885
+3-element Vector{Float64}:
+ 0.06373849439691504
+ 0.06373849439691504
+ 0.06373849439691504
 ```
 
 Normally, you can't use locally scoped variables in `@benchmark` or `@benchmarkable`, since all benchmarks are defined at the top-level scope by design. However, you can work around this by interpolating local variables into the benchmark expression:
@@ -228,17 +222,16 @@ julia> b = @benchmarkable sort!(y) setup=(y = copy($x))
 Benchmark(evals=1, seconds=5.0, samples=10000)
 
 julia> run(b)
-BenchmarkTools.Trial:
-  memory estimate:  0 bytes
-  allocs estimate:  0
-  --------------
-  minimum time:     5.739 ms (0.00% GC)
-  median time:      5.757 ms (0.00% GC)
-  mean time:        5.871 ms (0.00% GC)
-  maximum time:     62.138 ms (0.00% GC)
-  --------------
-  samples:          805
-  evals/sample:     1
+BechmarkTools.Trial: 819 samples with 1 evaluations.
+ Range (min … max):  5.983 ms …  6.954 ms  ┊ GC (min … max): 0.00% … 0.00%
+ Time  (median):     6.019 ms              ┊ GC (median):    0.00%
+ Time  (mean ± σ):   6.029 ms ± 46.222 μs  ┊ GC (mean ± σ):  0.00% ± 0.00%
+
+        ▃▂▂▄█▄▂▃                                                  
+  ▂▃▃▄▆▅████████▇▆▆▅▄▄▄▅▆▄▃▄▅▄▃▂▃▃▃▂▂▃▁▂▂▂▁▂▂▂▂▂▂▁▁▁▁▂▂▁▁▁▂▂▁▁▂▁▁▂
+  5.98 ms           Histogram: frequency by time           6.18 ms (top 1%)
+
+ Memory estimate: 0 bytes, allocs estimate: 0.
 ```
 
 In the above example, we wish to benchmark Julia's in-place sorting method. Without a setup phase, we'd have to either allocate a new input vector for each sample (such that the allocation time would pollute our results) or use the same input vector every sample (such that all samples but the first would benchmark the wrong thing - sorting an already sorted vector). The setup phase solves the problem by allowing us to do some work that can be utilized by the core expression, without that work being erroneously included in our performance results.
@@ -250,35 +243,32 @@ Note that the `setup` and `teardown` phases are **executed for each sample, not 
 It's possible for LLVM and Julia's compiler to perform optimizations on `@benchmarkable` expressions. In some cases, these optimizations can elide a computation altogether, resulting in unexpectedly "fast" benchmarks. For example, the following expression is non-allocating:
 ```julia
 julia> @benchmark (view(a, 1:2, 1:2); 1) setup=(a = rand(3, 3))
-BenchmarkTools.Trial:
-  memory estimate:  0 bytes
-  allocs estimate:  0
-  --------------
-  minimum time:     2.293 ns (0.00% GC)
-  median time:      2.302 ns (0.00% GC)
-  mean time:        2.330 ns (0.00% GC)
-  maximum time:     6.441 ns (0.00% GC)
-  --------------
-  samples:          10000
-  evals/sample:     1000
+BechmarkTools.Trial: 10000 samples with 1000 evaluations.
+ Range (min … max):  2.885 ns … 14.797 ns  ┊ GC (min … max): 0.00% … 0.00%
+ Time  (median):     2.895 ns              ┊ GC (median):    0.00%
+ Time  (mean ± σ):   3.320 ns ±  0.909 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%
 
+  █             ▁   ▁ ▁▁▁                                     ▂▃▃▁
+  █▁▁▇█▇▆█▇████████████████▇█▇█▇▇▇▇█▇█▇▅▅▄▁▁▁▁▄▃▁▃▃▁▄▃▁▄▁▃▅▅██████
+  2.88 ns        Histogram: log(frequency) by time         5.79 ns (top 1%)
+
+ Memory estimate: 0 bytes, allocs estimate: 0.0
 ```
 
 Note, however, that this does not mean that `view(a, 1:2, 1:2)` is non-allocating:
 
 ```julia
 julia> @benchmark view(a, 1:2, 1:2) setup=(a = rand(3, 3))
-BenchmarkTools.Trial:
-  memory estimate:  64 bytes
-  allocs estimate:  1
-  --------------
-  minimum time:     15.613 ns (0.00% GC)
-  median time:      17.825 ns (0.00% GC)
-  mean time:        23.358 ns (17.46% GC)
-  maximum time:     1.725 μs (95.12% GC)
-  --------------
-  samples:          10000
-  evals/sample:     998
+BechmarkTools.Trial: 10000 samples with 1000 evaluations.
+ Range (min … max):  3.175 ns … 18.314 ns  ┊ GC (min … max): 0.00% … 0.00%
+ Time  (median):     3.176 ns              ┊ GC (median):    0.00%
+ Time  (mean ± σ):   3.262 ns ±  0.882 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%
+
+  █                                                               
+  █▁▂▁▁▁▂▁▂▁▂▁▁▂▁▁▂▂▂▂▂▂▁▁▂▁▁▂▁▁▁▂▂▁▁▁▂▁▂▂▁▂▁▁▂▂▂▁▂▂▂▂▂▂▂▂▂▂▂▁▂▂▁▂
+  3.18 ns           Histogram: frequency by time           4.78 ns (top 1%)
+
+ Memory estimate: 0 bytes, allocs estimate: 0.8
 ```
 
 The key point here is that these two benchmarks measure different things, even though their code is similar. In the first example, Julia was able to optimize away `view(a, 1:2, 1:2)` because it could prove that the value wasn't being returned and `a` wasn't being mutated. In the second example, the optimization is not performed because `view(a, 1:2, 1:2)` is a return value of the benchmark expression.
@@ -317,66 +307,72 @@ This section provides a limited number of examples demonstrating these types. Fo
 Running a benchmark produces an instance of the `Trial` type:
 
 ```julia
-julia> t = @benchmark eig(rand(10, 10))
-BenchmarkTools.Trial:
-  memory estimate:  9.30 KiB
-  allocs estimate:  28
-  --------------
-  minimum time:     33.262 μs (0.00% GC)
-  median time:      38.618 μs (0.00% GC)
-  mean time:        39.981 μs (2.65% GC)
-  maximum time:     2.814 ms (95.07% GC)
-  --------------
-  samples:          10000
-  evals/sample:     1
+julia> t = @benchmark eigen(rand(10, 10))
+BechmarkTools.Trial: 10000 samples with 1 evaluations.
+ Range (min … max):  26.549 μs …  1.503 ms  ┊ GC (min … max): 0.00% … 93.21%
+ Time  (median):     30.818 μs              ┊ GC (median):    0.00%
+ Time  (mean ± σ):   31.777 μs ± 25.161 μs  ┊ GC (mean ± σ):  1.31% ±  1.63%
+
+             ▂▃▅▆█▇▇▆▆▄▄▃▁▁                                        
+  ▁▁▁▁▁▁▂▃▄▆████████████████▆▆▅▅▄▄▃▃▃▂▂▂▂▂▂▁▂▁▂▂▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁
+  26.5 μs           Histogram: frequency by time            41.3 μs (top 1%)
+
+ Memory estimate: 16.36 KiB, allocs estimate: 19.
 
 julia> dump(t) # here's what's actually stored in a Trial
-  BenchmarkTools.Trial
-    params: BenchmarkTools.Parameters # Trials store the parameters of their parent process
-      seconds: Float64 5.0
-      samples: Int64 10000
-      evals: Int64 1
-      overhead: Float64 0.0
-      gctrial: Bool true
-      gcsample: Bool false
-      time_tolerance: Float64 0.05
-      memory_tolerance: Float64 0.01
-    times: Array{Float64}((10000,)) [33262.0, 33793.0, … 2.77342e6, 2.81368e6] # every sample is stored in the Trial
-    gctimes: Array{Float64}((10000,)) [0.0, 0.0, … 2.66614e6, 2.67486e6]
-    memory: Int64 9520
-    allocs: Int64 28
+BenchmarkTools.Trial
+  params: BenchmarkTools.Parameters
+    seconds: Float64 5.0
+    samples: Int64 10000
+    evals: Int64 1
+    overhead: Float64 0.0
+    gctrial: Bool true
+    gcsample: Bool false
+    time_tolerance: Float64 0.05
+    memory_tolerance: Float64 0.01
+  times: Array{Float64}((10000,)) [26549.0, 26960.0, 27030.0, 27171.0, 27211.0, 27261.0, 27270.0, 27311.0, 27311.0, 27321.0  …  55383.0, 55934.0, 58649.0, 62847.0, 68547.0, 75761.0, 247081.0, 1.421718e6, 1.488322e6, 1.50329e6]
+  gctimes: Array{Float64}((10000,)) [0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0  …  0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 1.366184e6, 1.389518e6, 1.40116e6]
+  memory: Int64 16752
+  allocs: Int64 19
 ```
 
-As you can see from the above, a couple of different timing estimates are pretty-printed with the `Trial`. You can calculate these estimates yourself using the `minimum`, `median`, `mean`, and `maximum` functions:
+As you can see from the above, a couple of different timing estimates are pretty-printed with the `Trial`. You can calculate these estimates yourself using the `minimum`, `median`, `mean`, `maximum`, and `std` functions:
 
 ```julia
 julia> minimum(t)
-BenchmarkTools.TrialEstimate:
-  time:             33.262 μs
+BenchmarkTools.TrialEstimate: 
+  time:             26.549 μs
   gctime:           0.000 ns (0.00%)
-  memory:           9.30 KiB
-  allocs:           28
+  memory:           16.36 KiB
+  allocs:           19
 
 julia> median(t)
-BenchmarkTools.TrialEstimate:
-  time:             38.618 μs
+BenchmarkTools.TrialEstimate: 
+  time:             30.818 μs
   gctime:           0.000 ns (0.00%)
-  memory:           9.30 KiB
-  allocs:           28
+  memory:           16.36 KiB
+  allocs:           19
 
 julia> mean(t)
-BenchmarkTools.TrialEstimate:
-  time:             39.981 μs
-  gctime:           1.058 μs (2.65%)
-  memory:           9.30 KiB
-  allocs:           28
+BenchmarkTools.TrialEstimate: 
+  time:             31.777 μs
+  gctime:           415.686 ns (1.31%)
+  memory:           16.36 KiB
+  allocs:           19
 
 julia> maximum(t)
-BenchmarkTools.TrialEstimate:
-  time:             2.814 ms
-  gctime:           2.675 ms (95.07%)
-  memory:           9.30 KiB
-  allocs:           28
+BenchmarkTools.TrialEstimate: 
+  time:             1.503 ms
+  gctime:           1.401 ms (93.21%)
+  memory:           16.36 KiB
+  allocs:           19
+
+julia> std(t)
+BenchmarkTools.TrialEstimate: 
+  time:             25.161 μs
+  gctime:           23.999 μs (95.38%)
+  memory:           16.36 KiB
+  allocs:           19
 ```
 
 ### Which estimator should I use?
@@ -412,7 +408,7 @@ Calling the `ratio` function on two `TrialEstimate` instances compares their fie
 ```julia
 julia> using BenchmarkTools
 
-julia> b = @benchmarkable eig(rand(10, 10));
+julia> b = @benchmarkable eigen(rand(10, 10));
 
 julia> tune!(b);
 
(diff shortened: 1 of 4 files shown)


mcabbott:
This looks neat and seems easy to read. I like the colour showing mean/median, and I like that the units are clear at a glance. Going horizontally gets enough bins without making you scroll too much.

I have suggestions. Perhaps printing "Histogram: frequency by time" every time is a bit much. Perhaps printing the min/max twice is also unnecessary. It would be nicer to have mean & median on an equal footing, instead of median being tacked on at the right somewhere. So I'd propose that it look like this:
```
BechmarkTools.Trial:
 10000 samples with 968 evaulations.
 Time: median:     90.902 ns               (0.00% GC)
       mean ± σ:   94.936 ns ±  47.797 ns  (2.78% ±  5.03% GC)

          ▁▃▅▆▇█▇▆▅▂▁                                          
  ▂▂▃▃▄▅▆▇███████████▇▆▄▄▃▃▂▂▂▂▂▂▂▂▂▂▂▁▂▁▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂
  min: 77.7 ns (0.00% GC)                max: 137 ns (87.94% GC)

 Memory estimate: 160 bytes, allocations estimate: 1.
```
If you think it's insufficiently obvious that min/max are the left/right-most blocks, then these could be coloured, but otherwise could just be plain: perhaps grayscale plus two colours is enough.

tecosaur:
Thanks for the feedback @mcabbott !

> Perhaps printing "Histogram: frequency by time" every time is a bit much

The reason why I have this is because it's not always "frequency by time". From the suite of expressions I benchmarked I found cases in a frequency vs. time plot distribution was not very helpful. So, conditional on a little heuristic sometimes "log(frequency) by time" is used. If you have a suggestion that is more succinct but still captures this that would be great.

> Perhaps printing the min/max twice is also unnecessary

The max isn't printed twice. Because long tails seem common, but are not indicative of the performance characteristics of the expression I chop off the top 1% of times when creating the histogram. You can see this in the screenshot where the max time is ~710ns but the max on the graph is 140ns. Not doing this causes what I often consider to be the region of interest to become undesirably squished.

> It would be nicer to have mean & median on an equal footing, instead of median being tacked on at the right somewhere

I think the colouring gives it the same prominence, but this seems reasonable.

I hope this makes the motivation behind some of my choices a bit more clear :)

mcabbott:
Ah, if it's sometimes log, then perhaps that needs to be noted. Can you add an example above maybe?

I guess my complaint there is about saying "time" twice, and labelling a histogram "histogram" every time. But here's a possible idea; it doesn't seem too crammed to put min & max on the same line although I agree that 6 things would probably be too much. (I picture `[]` in light grey, as you had it.)
```
  ▂▂▃▃▄▅▆▇███████████▇▆▄▄▃▃▂▂▂▂▂▂▂▂▂▂▂▁▂▁▂▂▂▂▂▂▂▂▂▂▂▂▂▂ ... ▄
                               [linear axis, to 137 ns]
  min: 77.7 ns (0.00% GC)             max: 954.823 ns (87.94% GC)
```
My main complaint about the existing printing is that I never remember what order min/max/median/mean get printed; I think `mean ± σ` would help, and moving min/max would make them super-obvious.

tecosaur:
I've just realised that your approach would be misleading, as min/max GC don't necessarily correspond to min/max time :slightly_frowning_face:.

tecosaur:
Are any of the JuliaCI people planning on taking a look at this? Perhaps @Roger-luo, @vtjnash, or @vchuravy ?

tecosaur:
Ah oops, I assumed that Roger-luo and vtjnash were part of JuliaCI based on recent commits, but I see that that isn't the case. Sorry for the bogus ping.

mcabbott:
> min/max GC don't necessarily correspond to min/max time

This is a question I meant to raise too. What exactly is that GC percentage printed next to `mean`? There are several possibilities, is one more useful than the others? I would argue that the current printing certainly suggests that the GC percentages for min/max/median belong to the same runs as the times shown:
```
  minimum time:     248.763 ns (0.00% GC)
  median time:      262.463 ns (0.00% GC)
  mean time:        370.344 ns (28.55% GC)
  maximum time:     14.132 μs (97.67% GC)
```
If this isn't now the case, perhaps it should be... or perhaps the printing should separate these if in fact it's the median GC time, as a percentage (or median GC percentage?) not the GC percentage of the median-time run. (I don't see anything about this in the documentation.)

tecosaur:
Good point about it not being clear. How about this?

image

> I don't see anything about this point in the documentation.

I don't think it's documented, but I noticed when working with this that the time and GC results lists are both sorted.

mcabbott:
> but I noticed when working with this that the time and GC results lists are both sorted

That would usually be true for all the options I mentioned, right?

tecosaur:
> That would usually be true for all the options I mentioned, right?

What do you mean?

tecosaur:
Pushed the change in the screenshot above.

Arkoniak:
Before this PR is merged, can I ask you please to take a pause and think it through again?

In my opinion, such visual change is a big thing and shouldn't be taken lightly, even despite the best intentions, for accessibility reasons.

First of all, you are using lots of colors. While colors can be very expressive, at the same time they can be very distracting. The good thing about the current design that it is color neutral, which means that while it doesn't give you the best possible visual it is not offending anyone.

Secondly, there were issues when various colorschemes were broken when extra colors were introduced in logs in Julia 1.6 for example (https://github.com/JuliaLang/julia/issues/38730). One can argue that it is a problem of the colorscheme itself, but there were also issues when people had problems since they had problems with their eyes or monitors. 

I think, there can be multiple ways to solve this problem. These solutions can be for example:

1. Add a keyword `colorized` to `@benchmark` macro with default value `false`. 
2. Instead of hardcoded values use some sort of colorscheme (`:median_color` instead of `:blue`) and a dictionary, where one can setup any sort of color he/she wants. By default all colors should be `:white`
3. Instead of overriding `@benchmark` macro add `@colorized_hyperfine`.
4. Use white colors by default, but detect that `OhMyREPL.jl` is loaded and use colors in the latter case. 

For color schemes or `colorized` variant, one can add a possibility to set them globally through options or global environment variables, so those who wants to see colored version can add necessary lines in `startup.jl`.

Just to clarify, it is nice to have good looking output and it should be kept as a feature, but it should be opt-in. Colors shouldn't be a distraction or an obstacle to fulfilling the main purpose of `BenchmarkTools.jl`, i.e. benchmark code.

tecosaur:
@Arkoniak I appreciate your concern, but the only colours used here are 8-color terminal code. These are set by the terminal/user and should be readable at all times. My take is that this is the responsibility of the terminal/user. Should I be using 256-color codes, I would be inclined to agree that the use of colour would warrant more thought. This is why after trying a shaded gradient with the histogram I reverted to solid 'white'. The lone exception to this here is `:light_black`, but that seems to be addressed by `Base.text_colors[:light_black] = Base.text_colors[:white]` should it be an issue.

How "distracting" you want terminal colours to be is indeed a personal preference, and it can be addressed by users changing the base terminal colours to match their preference --- as they already do.

crstnbr:
I like colors if they are (1) customizable and (2) highlighting the most important things. In that regard I would ask the question: do we really need to highlight mean median min and max, i.e. all of them? Afaik, min and median are the most important ones, so why not just highlight those with colors? The full information is still there. One might argue that If you highlight / color everything you don't highlight anything anymore.

tecosaur:
The colours here aren't currently customisable, but if there's much demand I can make them so.
Regading
> The full information is still there. One might argue that If you highlight / color everything you don't highlight anything anymore.

I think there's an important point to be made here. If you *emphasise* everything you emphasise nothing, but colour can also be used for categorisation, as I do here. Which number corresponds to `max`? The one that's the same colour. That's not emphasis, that's categorisation. If you're looking at multiple benchmarks and wanting to compare a particular parameter, using different colours makes it easier to visually jump to what you're looking for. Hence, IMO the current level of colouring is warranted.

I think it's also worth bearing in mind that once/if this PR is accepted, it's not as if this feature is locked in stone. Tweaks (like colour customisation) can always be added later.

rafaqz:
The colors are great. But light_black should be reserved for text that carries no important information, or a lot of people wont be able to see it.

tecosaur:
Hmmm, maybe I should just make the GC text white then

rafaqz:
And probably the timings underneath each end of the histogram...

I just did the colorsheme in DimensionalData.jl and light black does look good, but I only used it for additional comments, brakcets etc that wont be missed if you cant see them. The diver, probaly "Histogram: frequency by time" is also non-essential, people will figure that out.

tecosaur:
That seems like a pretty reasonable approach, I'll give that a shot :+1:

tecosaur:
Implemented :slightly_smiling_face: 

image

rafaqz:
Looks great! But there's still that one dark `80.3 ns` ;)

I'm using this already it's a vast improvement for benchmarking. Thanks.

rafaqz:
IDK, but when I read it thats the value I read, not the one in the middle

tecosaur:
Fair enough. Here's a variant then which does that and applies grey to the parents and colons:

image

crstnbr:
This is how things look for me (I was wondering because @tecosaur's `@benchmark` is purple rather than orange):

Screenshot 2021-06-09 at 11 39 25

(I'd think that my terminal color scheme is pretty "standard", i.e. no drastic color changes.)

tecosaur:
Thanks @crstnbr. It's nice to see how things look with other people. The purple is just because of the OhMyREPL.jl theme I've set. I think I'm pretty happy with this current state of affairs. Now if I could just get a maintainer to take a look at this... 😔

fredrikekre:
Perhaps it would be worth keeping the old format around for the 2-arg signature (`show::IO, ::Trial)`)? Then you will still get this pretty printing in e.g. the REPL, but if you `print` or string interpolate the variable you get the old output.

pfitzseb:
What do you think about keeping the text ordering roughly the same, so that `Range` is the first entry? That's usually the most important one when only glancing at the output, no?

tecosaur:
Hmmm. I'm not sure. I'd think that the median value would be most indicative of the expected performance and so deserve to be first, but I'm open to other thoughts on this.

crstnbr:
According to this paper (from members of our community) the minimum is the most robust estimate in a noisy environment. This is also the reason why `@btime` prints the minimum rather than the median (or mean).

+1 To having the `Range` first as this will make the minimum the first number to read.

crstnbr:
One more thing, we might consider special casing IJulia/Jupyter (`MIME"text/html"`?).

Screenshot 2021-06-09 at 15 55 23

IMHO, in particular the yellow doesn't look as nice as in the terminal with a dark background.

tecosaur:
Ah cool, thanks for providing that info. In that case, how about this?

image

I've also tried colouring the minimum value bar for fun.

crstnbr:
I also noted that apart from the colouring some of the numbers are also bold (very much visible in the Jupyter notebook screenshot). Isn't the colouring enough?

Note that this also leads to the number of samples and evaluations (which is probably rather uninteresting information) to be most visible [in my terminal](https://github.com/JuliaCI/BenchmarkTools.jl/pull/217#issuecomment-857547978), because they are highlighted in white (best contrast) and bold.

crstnbr:
> I've also tried colouring the minimum value bar for fun.

In that case you should also highlight the maximum value bar (if visible) and probably also the `81.9 ns` below the minimum value bar. But frankly, I still think that less is more and, if anything, I would remove some of the colouring. Currently we are using 9(!) color/bold combos (white, white+bold, cyan+bold, purple, blue+bold, green+bold, green, grey, yellow).

pfitzseb:
`min`/`media` are probably the most important ones metrics, so it might make sense to only emphasize those?

mcabbott:
Note that the minimum time is written twice. I still think making min..max visually "rhyme" with mean+-std seems wrong. It prioritises making a tidy block of text over making the numbers easy to find. And that a more minimal use of colour would be clearer -- note the visual puzzle above of figuring out which shade of blue-ish text corresponds to the blue-ish graph bar, especially on the light background. Only one of the green numbers corresponds to the green bar.

I also ask again: what precisely are the GC percentages shown for median, mean, etc? Does anyone know without reading the source code?

tecosaur:
>  And that a more minimal use of colour would be clearer -- note the visual puzzle above of figuring out which shade of blue-ish text corresponds to the blue-ish graph bar, especially on the light background.

Only one blue makes sense. Bars in the middle clearly aren't the min, max, or standard deviation. You only need to be able to tell between median/mean which are blue/green.

> I also ask again: what precisely are the GC percentages shown for median, mean, etc?

Minimum GC %, maximum GC %, median GC %, ...

> IMHO, in particular the yellow doesn't look as nice as in the terminal with a dark background

Any suggestions? I think it should be something different to the surrounding text and not re-use one of the earlier colours.

> Also, the histogram seems to be cut by a white line.

Can't do anything about that, it's the notebook viewer's line spacing.

mcabbott:
> median GC %, .

So is it the GC percentage of the median-time run? Or the median GC time, as a percentage of the median time run? Or the median of the GC percentages of all runs?

crstnbr:
Incorporating some of the suggestions (min & median are the most important, less colors/bold, drop the range etc.) I made a more minimal draft. What do you guys think of this?

Screenshot 2021-06-09 at 18 10 45

ericphanson:
to me that looks more simple and clear. What about without any colors? They don't really look like they are adding much and the purple "Median" looks a little hard to read with the black background there. Maybe a vertical dashed line can indicate the median, or like textured blocks or something.

pfitzseb:
If we want to go with a less colorful option, how about bolding the minimum time and only having the median bar colored?

cmcaine:
If throughput is important to you, the mean time is the most relevant statistic. I think that might be a common enough case to be shown in the default view, and the addition of the std-dev to indicate dispersion is welcome to me.

Obviously it's easy enough to calculate your own summary statistics if you do care about that, but it might be nice to see them by default (tho the obvious risk is that people who don't know much about benchmarking will then use it more than they should).

It sounds like the algorithm for drawing the histogram will be fine, but I think it's probably worth checking how it looks when the time distribution is multimodal:

```julia
f(x) = x < 0 ? "neg" : string(x)
@benchmark f(x) setup=(x = randn()
```

(This is trimodal on my machine. I don't understand why it's not just bimodal :D)

chriselrod:
I agree. I think the mean is probably the most useful stat in allocating code.

I can reproduce the trimodality, using `BenchmarkHistograms.jl`:
```julia
julia> f(x) = x < 0 ? "neg" : string(x)
f (generic function with 1 method)

julia> b = @benchmark f(x) setup=(x = randn())
samples: 10000; evals/sample: 1000; memory estimate: 0 bytes; allocs estimate: 0
ns

 (1.0   - 13.0   ]  ██████████████████████████████ 5022
 (13.0  - 25.0   ]   0
 (25.0  - 37.0   ]   0
 (37.0  - 49.0   ]   0
 (49.0  - 61.0   ]   0
 (61.0  - 73.0   ]  ███▊616
 (73.0  - 85.0   ]  ████▋762
 (85.0  - 97.0   ]  ▏17
 (97.0  - 108.0  ]  ▏4
 (108.0 - 120.0  ]  ▏3
 (120.0 - 132.0  ]  ▋103
 (132.0 - 144.0  ]  ██▉480
 (144.0 - 156.0  ]  █████████████████▊2952
 (156.0 - 168.0  ]  ▎31
 (168.0 - 22366.0]  ▏10

                  Counts

min: 1.378 ns (0.00% GC); mean: 83.629 ns (23.47% GC); median: 1.407 ns (0.00% GC); max: 22.366 μs (99.43% GC).

julia> count(>(0), b.trial.gctimes)
9
```
Only 9 trials had more than 0 GC time, so that doesn't explain the large number of observations with the third mode.

tecosaur:
@cmcaine here's what I get with your test.

image

cmcaine:
Looks like your reporting of the median is off? If the histogram is from 1.7ns to 172ns then 79ns should be in the middle, not the far left.

Histogram looks good, tho, thanks.

crstnbr:
@tecosaur As suggested by others and illustrated in my draft above, I think it is much clearer to make the `Range` two separate points (minimum and maximum). We can then drop the `Time` bits for the median and mean and also drop all the GC specifications, i.e. `(min ... max)`, `(median)`, and `(mean ± σ)` since every line then clearly corresponds to min, max, median and mean, respectively.

MasonProtter:
Just my 2 cents: I love this the way it is. Having a colourless mode or a way to customize colours would be good for people with weird configs, but I really like the format and information layout.

tecosaur:
Given it's been almost a month and I haven't heard a peep from any maintainers, I'll likely hard-fork this and close the PR in a week or two. There's other stuff I want to try, and I'm already juggling PRs on other projects.

chriselrod:
> Given it's been almost a month and I haven't heard a peep from any maintainers, I'll likely hard-fork this and close the PR in a week or two. There's other stuff I want to try, and I'm already juggling PRs on other projects.

What about at BenchmarkHistograms?

tecosaur:
BenchmarkHistograms is doing its own thing with UnicodePlots, I think it's best left as separate. It's doing quantitative histograms with each bit labelled, this is a less clearly quantitative measure of the distribution. Besides which, the "other stuff" which I mentioned isn't just histograms, and from what I can tell BenchmarkHistograms is BenchmarkTools + Histograms.

chriselrod:
> BenchmarkHistograms is doing its own thing with UnicodePlots, I think it's best left as separate.

It doesn't depend on UnicodePlots. But you're right. Visually, this PR looks nicer to me and is more compact, but less informative.

ericphanson:
As I hinted at on Zulip, I'd be happy to put this in BenchmarkHistrograms (as long as @chriselrod likes the formatting lol, I think he's the only user I know of). I wanted to do a horizontal one originally but figured anything was better than nothing and didn't have much time to spend on it.

chriselrod:
I like it. I'd be happy for BenchmarkHistograms to go with this.

> (as long as @chriselrod likes the formatting lol, I think he's the only user I know of).

That's the main reason it'd be nice to get something here. I think people are slow to adopt.
But, I think adoption would be faster with this implementation. It's nice visually, and being relatively compact avoids the main downside of BenchmarkHistograms.
For reference if BenchmarkHistograms looks like this currently:
```julia
samples: 9097; evals/sample: 1; memory estimate: 0 bytes; allocs estimate: 0
ns

 (539000.0 - 540800.0]  ▏4
 (540800.0 - 542600.0]  █▊151
 (542600.0 - 544400.0]  █████████████▏1206
 (544400.0 - 546200.0]  ██████████████████████████████ 2760
 (546200.0 - 548000.0]  █████████████████████████████▎2682
 (548000.0 - 549800.0]  ██████████████▌1332
 (549800.0 - 551600.0]  █████▍488
 (551600.0 - 553400.0]  █▊157
 (553400.0 - 555300.0]  █▏94
 (555300.0 - 557100.0]  ▊67
 (557100.0 - 558900.0]  ▋54
 (558900.0 - 560700.0]  ▋50
 (560700.0 - 562500.0]  ▍34
 (562500.0 - 564300.0]  ▏8
 (564300.0 - 579200.0]  ▏10

                  Counts

min: 539.000 μs (0.00% GC); mean: 546.920 μs (0.00% GC); median: 546.458 μs (0.00% GC); max: 579.208 μs (0.00% GC).
```

ericphanson:
BenchmarkHistograms is not an alternative or competitor to BenchmarkTools; it automatically re-exports all of BenchmarkTools (the source code is very simple: https://github.com/ericphanson/BenchmarkHistograms.jl/blob/main/src/BenchmarkHistograms.jl). It's just a light layer on top to wrap the `@benchmark` macro with a visual show method. It's meant to be compositional.

Maybe that's what you you don't like and mean by not self-contained? But I think something like this is a way better solution than a fork, because any improvements or changes to the benchmarking only have one place to go (BenchmarkTools.jl) and are automatically inherited by BenchmarkHistograms (since it just re-exports them). With a fork, either you need to convince everyone to make PRs/updates to your fork instead of BenchmarkTools, or you constantly need to do maintenence to keep your fork up to date.

I don't really see what the issue is with the name; it's BenchmarkTools + a histogram, which is shortened to BenchmarkHistograms.

chriselrod:
If you're going to create a new library that doesn't depend on BenchmarkTools, I'd strongly recommend taking advantage of the opportunity to make some changes to the implementation.

For example, use closures instead of the `Expr`/ast copying behavior.
```julia
 function mybench(f::F, N) where {F}
    t = time_ns()
    for n ∈ 1:N
        f()
    end
    (time_ns() - t) / N
end
let x = Ref(1.2);
    mybench(100_000_000) do
        sin(x[])
    end
end
```
This is just meant to show something without all the `:copyast`, etc:
```julia
julia> @macroexpand @benchmark f!($du, $u0, nothing, 0.0)
quote
    #= /home/chriselrod/.julia/packages/BenchmarkHistograms/HvH8w/src/BenchmarkHistograms.jl:73 =#
    (BenchmarkHistograms.BenchmarkHistogram)(begin
            #= /home/chriselrod/.julia/packages/BenchmarkTools/MPPDO/src/execution.jl:386 =#
            local var"##257" = begin
                        #= /home/chriselrod/.julia/packages/BenchmarkTools/MPPDO/src/execution.jl:440 =#
                        BenchmarkTools.generate_benchmark_definition(Main, Symbol[], [Symbol("##du#258"), Symbol("##u0#259")], $(Expr(:copyast, :($(QuoteNode(:(f!(var"##du#258", var"##u0#259", nothing, 0.0))))))), Core._expr(:block, $(Expr(:copyast, :($(QuoteNode(nothing))))), Core._expr(:(=), Symbol("##du#258"), du), Core._expr(:(=), Symbol("##u0#259"), u0)), $(Expr(:copyast, :($(QuoteNode(nothing))))), BenchmarkTools.Parameters())
                    end
            #= /home/chriselrod/.julia/packages/BenchmarkTools/MPPDO/src/execution.jl:387 =#
            (BenchmarkTools).warmup(var"##257")
            #= /home/chriselrod/.julia/packages/BenchmarkTools/MPPDO/src/execution.jl:388 =#
            (BenchmarkTools).tune!(var"##257")
            #= /home/chriselrod/.julia/packages/BenchmarkTools/MPPDO/src/execution.jl:389 =#
            (BenchmarkTools).run(var"##257")
        end)
end
```
`$` could be used for adding those variables to a `let` block around the closure.

I'm not familiar with BenchmarkTool's internals, how it's doing things, why it's doing them, or if more minor changes can avoid the disadvantages of its approach.

But it does have disadvantages, most significantly that it leaks memory.
BLASBenchmarksCPU.jl, for example, has to allocate large vectors, and then use reshaped views of these vectors for all matrix sizes benchmarked. Every array used in a `BenchmarkTools` benchmark never gets freed, therefore it is forced to use the same array for each.

LoopVectorization's benchmarks are run in worker processes, and it must repeatedly `rmproc()` and `addproc()`, otherwise the leaked memory will total >20 GiB by the time it is done.

The `@macroexpand` is also fairly incomprehensible; I think there's an advantage to being more Julian.

If you're creating a separate library that doesn't depend on `BenchmarkTools`, I think it'd be great to drop the baggage.

tecosaur:
> BenchmarkHistograms is not an alternative or competitor to BenchmarkTools; it automatically re-exports all of BenchmarkTools (the source code is very simple: https://github.com/ericphanson/BenchmarkHistograms.jl/blob/main/src/BenchmarkHistograms.jl). It's just a light layer on top to wrap the `@benchmark` macro with a visual show method. It's meant to be compositional.

Some of the things I'm thinking of doing I don't think would be as cleanly compositional as this.

> Maybe that's what you you don't like and mean by not self-contained? But I think something like this is a way better solution than a fork, because any improvements or changes to the benchmarking only have one place to go (BenchmarkTools.jl) and are automatically inherited by BenchmarkHistograms (since it just re-exports them).
>
> With a fork, either you need to convince everyone to make PRs/updates to your fork instead of BenchmarkTools, or you constantly need to do maintenence to keep your fork up to date.

Yea, but pulling + merging is easy, and it doesn't look like this is that actively maintained so I'm thinking a fork would be a lot less mental effort than making a bunch of PRs here.

> I don't really see what the issue is with the name; it's BenchmarkTools + a histogram, which is shortened to BenchmarkHistograms.

Mmm. On this, I was thinking of adding more non-histogram things, which is why I'm not sure this may be the best direction to go in - e.g. statistical tests.

> If you're going to create a new library that doesn't depend on BenchmarkTools, I'd strongly recommend taking advantage of the opportunity to make some changes to the implementation.

The temptation to 'shake things up' and do exactly that is unfortunately steadily increasing for me... :laughing:

ericphanson:
> Some of the things I'm thinking of doing I don't think would be as cleanly compositional as this.

Ah ok, BenchmarkHistograms might not be a good fit then. I guess that applies to the name objection too. I really do want it just be a light wrapper; I don't want to be maintaining a benchmarking library myself 😄 .

tecosaur:
Cool :+1: 

> I don't want to be maintaining a benchmarking library myself

Frankly, neither do I :stuck_out_tongue:, but I'd much rather see what I'm thinking happen than not happen.

If by some chance I don't ever get around to adding the beyond-histograms stuff I've alluded to, I'd be happy to revisit the idea of contributing this to BenchmarkHistograms if BenchmarkTools is still unresponsive :slightly_smiling_face:.

DilumAluthge:
@ararslan @vtjnash @vchuravy @jrevels Could you weigh in on whether you think this PR can go into BenchmarkTools, or if it would be better suited in a separate package?

ararslan:
Hey @tecosaur, welcome and thanks for the contribution! Sorry for the lack of attention here.

I wouldn't consider myself an active maintainer here anymore, but I was tagged in so I'll opine. 😄 I really love the idea here, but I feel like for me personally, I would want to opt into this display rather than having it be the default from `@benchmark`. That could be as simple as a separate macro that implemments this, a flag to `@benchmark`, or what have you.

That's just my 2 cents, but I'll defer to the folks who actively maintain this package. 🙂

Roger-luo:
If people don't like the color can we configure that with Preference.jl or environment variables? I mean other things looks good to me. I'm curious what is holding this PR?

DilumAluthge:
> I'm curious what is holding this PR?

I think we need an active maintainer that's willing to accept this code, since they are the one that would be maintaining this code if it's added to this package.

DilumAluthge:
I talked with two of the active maintainers (Valentin and Jarrett) out-of-band. They both like the idea of this PR and will leave reviews tomorrow or the day after.

chriselrod:
Instead of a preference or a flag, why not make it a different macro?
BenchmarkTools already has `@btime`, `@belapsed`, `@benchmark` that each benchmark your code but have different outputs.
Why not another macro that displays a histogram?

rafaqz:
Maybe `@bhist` ?. 

Also, I really think this should be merged here. Its a big improvement to a very frequently used tool.

cmcaine:
I think defaults matter a lot and that this (or the simpler variant from a
few comments back) is a better default than the current ***@***.***`.

The original behaviour could be behind a flag, that way most people will
end up using the better presentation.

This is a slight preference, ***@***.***` is also good :)

jrevels (approved):
similar to @ararslan I'm not heavily involved with BenchmarkTools anymore, so this approval should probably be considered more of a "spiritual approval" 😁 

But I'm a big fan of this change, histograms are way more useful than a few summary stats for many benchmarks. I'm pro histograms-by-default and am in favor of having this be the default display option for `@benchmark`.

Otherwise, I like the thoughtful discussion over how to improve customizability options for this new feature, but I don't think it needs to block this feature getting merged in in the first place, as folks can always improve on it post-merge if we can align on what those improvements should look like 

Of course, I'll let @vchuravy weigh in more thoroughly, as he's been doing the most active maintenance of the package in recent times

src/trials.jl
```diff
@@ -289,6 +290,32 @@ function withtypename(f, io, t)
...
+
+function asciihist(bins, height=1)
+    histbars = ['▁', '▂', '▃', '▄', '▅', '▆', '▇', '█']
+    if minimum(bins) == 0
+        barheights = 2 .+ round.(Int, (height * length(histbars) - 2) * bins ./ maximum(bins))
+        barheights[bins .== 0] .= 1
+    else
+        barheights = 1 .+ round.(Int, (height * length(histbars) - 1) * bins ./ maximum(bins))
+    end
+    [let bheight = barheights[b] - (h-1) * length(histbars)
```
vchuravy, line 314: I find this form hard to read, can we use a `map` or explicit for loop instead?

src/trials.jl
```diff
@@ -313,36 +340,139 @@ Base.show(io::IO, t::TrialJudgement) = _show(io, t)
...
+        medtime, medgc = "N/A", "N/A"
+        avgtime, avggc = "N/A", "N/A"
+        stdtime, stdgc = "N/A", "N/A"
+        mintime, mingc = "N/A", "N/A"
+        maxtime, maxgc = "N/A", "N/A"
+        memorystr, allocsstr = "N/A", "N/A"
     end
-    println(io, "BenchmarkTools.Trial: ")
+
+    lmaxtimewidth = maximum(length.([medtime, avgtime, mintime]))
```
vchuravy, line 366:
```suggestion
    lmaxtimewidth = maximum(length.((medtime, avgtime, mintime)))
```

as well as below.

vchuravy: I might even prefer `max(length(medtime), length(avgtime), length(mintime))`

src/trials.jl
```diff
@@ -313,36 +340,139 @@ Base.show(io::IO, t::TrialJudgement) = _show(io, t)
...
+    printstyled(io, "┊"; color=:light_black)
+    print(io, " GC ")
+    printstyled(io, "("; color=:light_black)
+    print(io, "min … max")
+    printstyled(io, "): "; color=:light_black)
+    print(io, lpad(mingc, lmaxgcwidth), " … ", lpad(maxgc, rmaxgcwidth))
+
+    # Histogram
+
+    histtimes = t.times[1:round(Int, 0.99*end)]
```
vchuravy, line 428: What is the goal here? Is this supposed to be `t.times[1:end-1]`?

vchuravy (commented):
@tecosaur thank you for all the effort you have put into this sofar! I saw this earlier and was quite excited about it, but I didn't want to abridge the discussion around what the best presentation of the histogram is. 

I left some comments on the code inline. I am excited to merge this, but I would like to add a bit of configurability. I am not sure that making the interface more complicated by adding another macro or by adding keyword arguments, but in Cthulhu we have a global CONFIG object that can be used to turn of highlighting.

https://github.com/JuliaDebug/Cthulhu.jl/blob/886f6a709983dcb2406ad2f55881fc1dcdf5caac/src/Cthulhu.jl#L21-L37

Could you add something similar to that for allowing users to turn of color? (and maybe the histogram mode altogether)

src/trials.jl
tecosaur: Refactored to:
```jl
    heightmatrix = [min(length(histbars), barheights[b] - (h-1) * length(histbars))
                    for h in height:-1:1, b in 1:length(bins)]
    map(height -> if height < 1 ' ' else histbars[height] end, heightmatrix)
```

src/trials.jl
tecosaur: Done.

src/trials.jl
tecosaur: I found that without slicing off the top 1% of times anomalous results would dramatically increase the maximum end of the scale, squashing 99% of the data onto 10% of the scale (or similar). I think this is pretty undesirable.

Extracted to a helpfully named variable and annotation added to the scale for less "magic number"-ness.

tecosaur:
@vchuravy @jrevels thank you for getting back to me! As you can tell from the above, I was beginning to lose hope.

I've addressed the inline comments. Regarding the no-color option, adding http://no-color.org/ support to Julia's base print capabilities would feel like the best solution to me. This is a much bigger ask, but it would be a pain to switch out all the colours --- so I'd rather not if possible.

Lastly, with an eye to future contributions, should I expect a similar time frame / responsiveness?

StefanKarpinski:
Adding `NO_COLOR` support to Julia itself is a great idea. Do you want to file an issue and/or make a pull request?

StefanKarpinski:
Also, I love this PR. I'm a sucker for colors and nice terminal graphics, so the more the merrier where I'm concerned 🙂. I do think there's some good marketing value from just making this the default for BenchmarkTools: people will use it and others will see it and the reaction will be "wow, what is that?" with the answer "oh, this is just the standard way to benchmark stuff in Julia — it's really cool" and will convince people to try out the language. I'm pretty sure the nice REPL with colorized output and carefully elided arrays has convinced more than a few people to try out Julia.

tecosaur:
I've actually been trying to make a PR, but I've been unable to build Julia `master` with GCC 11 :frowning_face:.

This many not quite be the right approach, but this is what occurred to me (in `base/util.jl`):
```diff
@@ -71,7 +71,7 @@ function with_output_color(@nospecialize(f::Function), color::Union{Int, Symbol}
         bold::Bool = false, underline::Bool = false, blink::Bool = false,
         reverse::Bool = false, hidden::Bool = false)
     buf = IOBuffer()
-    iscolor = get(io, :color, false)::Bool
+    iscolor = get(io, :color, false)::Bool && ! "NO_COLOR" in keys(ENV)
```

StefanKarpinski:
> This many not quite be the right approach, but this is what occurred to me (in `base/util.jl`):

That does seem like it might be sufficient!

crstnbr:
Should probably be named `JULIA_NO_COLOR` right?

tecosaur:
@crstnbr if you look at https://no-color.org, `NO_COLOR` is what we should use.

StefanKarpinski:
> Should probably be named `JULIA_NO_COLOR` right?

No, because it's a common standard. The point is to have a single environment variable that can be set to disable color for all programs, not just Julia.

src/trials.jl
ericphanson: in BenchmarkHistograms I resolved this by using the 0.999 quantile of the data

> We choose a value Q = quantile(x, 0.999) to represent close-to-the-upper end of the data we might expect. Then, if Q is more than 2 bins away from the actual maximum, we collapse all of those bins into a single bin.

https://github.com/ericphanson/BenchmarkHistograms.jl/pull/13

tecosaur:
@StefanKarpinski  Great! I'm happy to make an untested PR, or leave it to someone else.

Oh and since I have a "main Julia person" who likes terminal colours here, I'd just like to express my *very strong* desire that there be a less painful way to use colours in Julia without pulling in packages:
```jl
    print(io, "\n", pad, " Time  ")
    printstyled(io, "("; color=:light_black)
    printstyled(io, "mean"; color=:green, bold=true)
    print(io, " ± ")
    printstyled(io, "σ"; color=:green)
    printstyled(io, "):   "; color=:light_black)
    printstyled(io, lpad(avgtime, lmaxtimewidth); color=:green, bold=true)
    print(io, " ± ")
    printstyled(io, lpad(stdtime, rmaxtimewidth); color=:green)
    ...
 ```
it would have been **much** nicer if this could have been something along the lines of:
```jl
using Base.color

print(io, "\n", pad, " Time ",
      light_black("("), green(bold("mean")), " ± ", green("σ"), light_black("):"),
      green(bold(lpad(avgtime, lmaxtimewidth))), " ± ", green(lpad(stdtime, rmaxtimewidth)) ...)
```

If there's any interest in this it's probably worth going to the Julia repo to discuss a good model for storing colour information. I may be biased by usage :stuck_out_tongue: but suspect it could be worth taking inspiration from the Emacs model of coloured strings being plain content + a set of ranges-style pairs. In my experience, this works rather seamlessly when composing text and modifying styles. I have an idea in my mind of what could be a rather nice re-implementation of this model, and would be happy to try and prototype something should this thought be of interest.

vchuravy:
@tecosaur can you fix the CI failures? I would then merge this :)

> Lastly, with an eye to future contributions, should I expect a similar time frame / responsiveness?

Hopefully faster, but I can't commit to anything. Feel free to ping me on slack if a PR get's stuck.

DilumAluthge:
@vchuravy @tecosaur Probably it would be good to also bump the minor version number too? Then you can register a new release as soon as you merge.

tecosaur:
It seems the issue `LoadError: syntax: space not allowed before "'"` is from
```jl
map(height -> if height < 1 ' ' else histbars[height] end, heightmatrix)
```
I'll try adding semicolons.

tecosaur:
Now I just need to figure out how to indicate the chopped off results. I considered bundling with the max bin, but that seems a tad misleading. Let me know what you think of these options

### A

image

### B

image

### C

image

### D

image

### E

Something else, let me know what you're thinking.

vchuravy:
C or E

StefanKarpinski:
C it is.

tecosaur:
From this \*cough _very large_ sample size, I feel tempted to say that the **C**s have it :stuck_out_tongue:. One last niggle: I'm having second thoughts about having this be red (I feel that it could be interpreted as "something went wrong" in this portion of the results), so I'm currently thinking **C**, but without the red.

Unrelated: Stefan, might you have any comments on my earlier thought? https://github.com/JuliaCI/BenchmarkTools.jl/pull/217#issuecomment-866756136

crstnbr:
FWIW, I'd go with C as well and +1 for not making it red. I feel it's too much of a signal color for this rather unimportant information.

vchuravy:
@tecosaur good to merge?

StefanKarpinski:
> Unrelated: Stefan, might you have any comments on my earlier thought? [#217 (comment)](https://github.com/JuliaCI/BenchmarkTools.jl/pull/217#issuecomment-866756136)

Having the ability to nest coloring/bold would be great. Jameson has an old PR that started on that which you may want to have a look at: https://github.com/JuliaLang/julia/pull/27430. The basic idea is to save color and boldness information in the `IOContext` object so that color and bold can be nested properly.

tecosaur:
@vchuravy I just implemented **C** without red, so unless there's anything else you'd like to see, I'd say this is ready to merge :slightly_smiling_face:.

DilumAluthge:
Would be good to get a new release.

StefanKarpinski:
The new output format is so nice! Bravo 👏🏻

tecosaur:
Glad you like it :grin: I love how Julia seems much more enthusiastic about making a rich terminal experience than most other languages I've used :heart:.

giordano:
> I've actually been trying to make a PR, but I've been unable to build Julia master with GCC 11 slightly_frowning_face.

Uh?  https://github.com/JuliaLang/julia/issues/40703 was fixed before you posted the message here.

tecosaur:
I saw, the fact of the matter is that I wasn't able to build it though, and I have all of the dependencies.

src/trials.jl
```diff
@@ -313,36 +340,151 @@ Base.show(io::IO, t::TrialJudgement) = _show(io, t)
...
+    if median(filter(b -> b > 0.1 * length(t.times) / histwidth, bins)) / maximum(bins) < 0.05
+        bins, logbins = log.(1 .+ bins), true
+    end
+    hist = asciihist(bins, histheight)
+    hist[end,end-1] = ' '
+    maxbin = maximum(bins)
+
+    delta1 = (histtimes[end] - histtimes[1]) / (histwidth - 1)
+    medpos = 1 + round(Int, (histtimes[length(histtimes) ÷ 2] - histtimes[1]) / delta1)
+    avgpos = 1 + round(Int, (mean(histtimes) - histtimes[1]) / delta1)
```
cmcaine, line 442: As mentioned earlier, the columns highlighted on the histogram are the median and the mean of the filtered times (the bottom 99% of times), which means that they may not match up with the text above, which shows the reported median and mean of all the times.

You can see this in my example from earlier:
2021-07-09-093910_790x292_scrot

Here, the median column clearly does not contain the median time.

Looks like a simple enough fix. If I have time today I will make a PR.

tecosaur: Cool, a fix PR would be nice :) Though since it's a log y-axis there I wouldn't be surprised if nothing changed.

cmcaine: It's the position on the x axis that's wrong. Unless I'm badly misunderstanding something it's definitely highlighting the wrong column

Image showing where I think it should go:

2021-07-09-093910_790x292_scrot-2

tecosaur: Yea, that's definitely not what it should be. The leftmost bar is the tallest, and so I'd expect the median to be there. Since this is a log-y being slightly taller is exponentially more points in that bin. The reason why I introduced a log mode at all is so that you could see if any other times where being hit more than others, as otherwise that single bar will completely dominate the plot.

tecosaur: Ah, ooops - I misread the numbers above ... somehow. I think you're completely right. Cary on :sweat_smile:

KristofferC:
So now we have had this for a while and I am curious what people think. Personally, I think the new output is really cool but the level of noise to signal is in most cases very high. This makes me doubt that having this as a default display is that good. Something terser that would fit on one line would be great, and then there could be an option to expand this to have the richer printing.

pfitzseb:
Isn't that basically `@btime` though? Or, rather, what part of the output is the signal for you? Something like
```
julia> b = @benchmark sort(data) setup=(data=rand(10))
BenchmarkTools.Trial: 10000 samples with 973 evaluations.
 67.764 ns, 0.00% GC (median 85.830 ns, 0.00% GC), 1 allocs for 160 bytes

julia> verbose(b)
BenchmarkTools.Trial: 10000 samples with 973 evaluations.
 Range (min … max):  67.764 ns … 793.284 ns  ┊ GC (min … max): 0.00% … 81.37%
 Time  (median):     85.830 ns               ┊ GC (median):    0.00%
 Time  (mean ± σ):   92.337 ns ±  34.990 ns  ┊ GC (mean ± σ):  1.83% ±  4.91%

    ▁▄▅▆▇███▇▇▆▅▄▄▄▂▂▁▁▁                                       ▃
  ▂▄██████████████████████▇██▇█▇▇▇▇▇▇▆▇▇▇▇▇▇▇███▇█▇▆▇▆▇▇▅▆▅▅▆▅ █
  67.8 ns       Histogram: log(frequency) by time       183 ns <

 Memory estimate: 160 bytes, allocs estimate: 1.
```

KristofferC:
Yes, something like that. There could fit be a +-σ there as well.

tecosaur:
I'm quite happy to say that the new histogram benchmark, while displaying more information, is actually more compact than the previous display to the point that it's actually less lines. Regarding,
> Something terser that would fit on one line would be great

It sounds like this is more an argument for expanding `@btime` than contracting `@benchmark`. It's currently just `time (allocations: bytes)` but you could suggest something like `time (min time-max time, allocations: bytes)` etc.

timholy:
I like the status quo. The histogram makes it crystal-clear what's happening, and gets people thinking more deeply about performance measurement.

I'd support an `io = IOContext(stdout, :compact=>true)` for a more compact display.

KristofferC:
> The histogram makes it crystal-clear what's happening, and gets people thinking more deeply about performance measurement.

In the vast majority of cases, it shows literally noise (your computer doing various background tasks which is a bad thing to show because you are likely to confuse the runtime of your code based on external "pollution".) I have yet seen a single comment about the histogram result, in fact, I have not seen a single comment about almost any of all the new numbers.
All I've seen from people doing e.g. presentations with the new output is them having trouble finding the relevant data in the sea of numbers. And I can't read benchmarks on slack anymore because the output is so wide that it line breaks and turn the ASCII art into garbage. 

Just as a reference, for people who are interested, https://youtu.be/vrfYLlR8X8k?t=915 (and about 7 minutes forward) is a bit of an interesting discussion of timing code from the guy who used to do a lot of performance work at facebook.

timholy:
So then why not just use `@btime`?

My question is often "is the difference in performance between two implementations meaningful, or noise?" That's the whole reason I just submitted #255, with a real-world demo in https://github.com/JuliaLang/julia/issues/42227#issuecomment-920609318. You could (correctly) make the argument that if I'm going to the effort to supply an `IOContext` then I could turn on a more verbose-mode printing. I don't have strong objections to that view, although elsewhere we typically are verbose unless `:compact => true`. Even without that pretty-printing across runs, often you can get by looking at the standard deviation (I would in fact propose we add the absolute deviation to the median). So yes, I do look at the numbers.

For reference where seeing the raw data can be helpful: https://stackoverflow.com/questions/69164027/unreliable-results-from-timeit-cache-issue which came up in working on https://github.com/JuliaImages/image_benchmarks. And the video is useful, thanks. I think it actually provides fodder for the view that we should show the histogram: he points out that it is a bit complicated, that there are circumstances like networking where the minimum is not the only measure you're interested in.

I should make it clear that I can live with any outcome; I'm mostly glad this is available, and if we switch to a less verbose printing I will turn it on manually.

StefanKarpinski:
I feel the same way as @timholy here. If brevity is so crucial, why not use `@btime`? The `@benchmark` macro has always been about getting more detailed information and the new version with histogram takes up fewer lines than the old version without it did.

Moelf:
if `@btime` can show allocation and compilation time and GC % when it's not 0, that would be enough.

Basically, I think the idea is sometimes people want communicable (via non-screenshot format, potentially in a limited width setting like slack thread) benchmark with real-life workload (which means having comparing allocation, GC time etc.) *without wrapping 30 lines with black-and-while boxes.

I'm all for keeping everything in `@benchmark` except re-organize them so the sender can selectively copy-paste without the need to manually edit different multiple lines, as outlined in https://github.com/JuliaCI/BenchmarkTools.jl/issues/236

timholy:
Not to beat a dead horse, but yesterday I got this histogram:
```
         █             ▃      ▃ ▃ ▃  ▃ ▃█    ▃ █ ▃ ▃  █▃▃█      
  ▇▁▁▁▁▁▇█▁▁▁▇▁▇▁▇▁▇▁▇▁█▁▁▇▇▇▁█▇█▇█▁▇█▁██▁▁▁▁█▇█▇█▇█▁▇████▇▁▁▇ ▁
  68 ms           Histogram: frequency by time          102 ms <
```
It instantly told me that I had some kind of "jackpot event" (of low quantal count and thus high variability) dominating the runtime, which in this case proved to be thread scheduling. Took about half a second to diagnose thanks to the histogram.

chriselrod:
I think histograms are worth looking at whenever using `Threads.@spawn` or `Threads.@threads`, e.g.
```julia
julia> @benchmark fetch(Threads.@spawn 1+1)
BenchmarkTools.Trial: 7468 samples with 10 evaluations.
 Range (min … max):  10.931 μs … 360.687 μs  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     66.520 μs               ┊ GC (median):    0.00%
 Time  (mean ± σ):   66.620 μs ±   5.501 μs  ┊ GC (mean ± σ):  0.00% ± 0.00%

                                                        ▅█▆▁
  ▂▁▂▂▁▁▂▁▁▁▁▁▂▂▁▁▁▁▁▁▁▁▂▁▁▁▁▁▁▂▂▁▂▁▂▁▁▂▂▁▂▂▁▁▂▁▁▁▁▂▂▂▄▇████▆▃ ▃
  10.9 μs         Histogram: frequency by time         70.8 μs <

 Memory estimate: 480 bytes, allocs estimate: 5.
```
The minimum time isn't representative at all.

mcabbott:
There is space on the line to make `@btime` print a bit more detail, maybe we should do that too? I'd argue that omitting the minimum would be weird, and that if you want one more time it should be the mean not the median --- they agree in this `@spawn` example, but disagree in cases with many fast & a few slower, like   https://github.com/JuliaCI/BenchmarkTools.jl/issues/237#issuecomment-922257927 .

chriselrod:
I agree. While with a single number summary the mean can be biased by a single slow run, given two numbers I think it's a better addition to the minimum than the median.

KristofferC:
> If brevity is so crucial, why not use @btime?

Unfortunately, I spend a lot more time reading others results than my own.thisbis my life now:

Screenshot_2021-09-22-20-08-40-617_com Slack

:p

ericphanson:
I think many folks used to report the `@btime` results before-- I don't remember seeing much of the old printing before the histograms. Post-histograms, folks seem to be reporting the results of `@benchmark` more.

So I feel like it's the fact that users switched which one they use to report results (probably because of the fancier printing) rather than what the default of `@benchmark` is.

timholy:
> This is my life now:

Agreed, that's really awful. But I agree with @ericphanson that the reason more people are using `@benchmark` now is because of the histogram. That certainly describes me. You still haven't answered why people can't just use `@btime`. Can we focus on that question? What would it take to make that acceptable advice? You could just say, "sorry, I can't read that. Please quote the results of `@btime`" and your suffering is over.

docs/src/manual.md
```diff
@@ -317,66 +307,72 @@ This section provides a limited number of examples demonstrating these types. Fo
...
+  gctime:           1.401 ms (93.21%)
+  memory:           16.36 KiB
+  allocs:           19
+
+julia> std(t)
+BenchmarkTools.TrialEstimate: 
+  time:             25.161 μs
+  gctime:           23.999 μs (95.38%)
+  memory:           16.36 KiB
+  allocs:           19
```
milesfrain, line 375: Cross-link to #146 for reexporting `std` too.

PR #243, opened 2021-08-03
Fix compatibility with Julia v1.0

vyu:
In Julia v1.0, the use of `eachrow` and `isnothing` breaks the printing of benchmark results as they were only added in v1.1. This PR replaces them with compatible alternatives.

### DIFF
--- a/src/trials.jl
+++ b/src/trials.jl
@@ -466,13 +466,7 @@ function Base.show(io::IO, ::MIME"text/plain", t::Trial)
         end
     end
 
-    function remtrailingzeros(timestr)
-        if match(r"\.0+$", timestr) !== nothing
-            replace(timestr, r"\.0+$" => "")
-        else
-            replace(timestr, r"(\.\d+?)0+$" => s"\1")
-        end
-    end
+    remtrailingzeros(timestr) = replace(timestr, r"\.?0+ " => " ")
     minhisttime, maxhisttime = remtrailingzeros.(prettytime.(round.(histtimes[[1; end]], sigdigits=3)))
 
     print(io, "\n", pad, "  ", minhisttime)
--- a/src/trials.jl
+++ b/src/trials.jl
@@ -456,9 +456,8 @@ function Base.show(io::IO, ::MIME"text/plain", t::Trial)
 
     print(io, "\n")
     for r in axes(hist, 1)
-        histrow = view(hist, r, :)
         print(io, "\n", pad, "  ")
-        for (i, bar) in enumerate(histrow)
+        for (i, bar) in enumerate(view(hist, r, :))
             color = :default
             if i == avgpos color = :green end
             if i == medpos color = :blue end
--- a/src/trials.jl
+++ b/src/trials.jl
@@ -455,7 +455,8 @@ function Base.show(io::IO, ::MIME"text/plain", t::Trial)
     end
 
     print(io, "\n")
-    for histrow in eachrow(hist)
+    for r in axes(hist, 1)
+        histrow = view(hist, r, :)
         print(io, "\n", pad, "  ")
         for (i, bar) in enumerate(histrow)
             color = :default
@@ -466,7 +467,7 @@ function Base.show(io::IO, ::MIME"text/plain", t::Trial)
     end
 
     function remtrailingzeros(timestr)
-        if ! isnothing(match(r"\.0+$", timestr))
+        if match(r"\.0+$", timestr) !== nothing
             replace(timestr, r"\.0+$" => "")
         else
             replace(timestr, r"(\.\d+?)0+$" => s"\1")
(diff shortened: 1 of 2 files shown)


vchuravy:
Thanks for your PR! Could you add a test?

src/trials.jl
```diff
@@ -455,7 +455,8 @@ function Base.show(io::IO, ::MIME"text/plain", t::Trial)
     end
 
     print(io, "\n")
-    for histrow in eachrow(hist)
+    for r in axes(hist, 1)
+        histrow = view(hist, r, :)
         print(io, "\n", pad, "  ")
         for (i, bar) in enumerate(histrow)
```
tecosaur, line 461: Why not replace `enumerate(histrow)` with `enumerate(hist[r,:])` ?

vyu:
Yes, sure, I've added a test on the exact output of `show(::IO, ::MIME"text/plain", t::Trial)` for a simple `t`.

The test found a bug in the [previous bugfix](https://github.com/JuliaCI/BenchmarkTools.jl/commit/ed0096a29825d02682d0f617ab934531528e5824) to `remtrailingzeros` in that it no longer removed trailing zeros (the anchor needs to be a space since `timestr` is of the form `\d+\.\d{3} \w+`), so I've fixed that as well.

test/TrialsTests.jl
```diff
@@ -233,4 +233,17 @@ else
 
 end
 
+trial = BenchmarkTools.Trial(BenchmarkTools.Parameters(), [1.0, 1.01], [0.0, 0.0], 0, 0)
```
tecosaur, line 236: Should this be its own commit?

src/trials.jl
```diff
@@ -465,13 +466,7 @@ function Base.show(io::IO, ::MIME"text/plain", t::Trial)
...
     end
 
-    function remtrailingzeros(timestr)
-        if ! isnothing(match(r"\.0+$", timestr))
-            replace(timestr, r"\.0+$" => "")
-        else
-            replace(timestr, r"(\.\d+?)0+$" => s"\1")
-        end
-    end
+    remtrailingzeros(timestr) = replace(timestr, r"\.?0+ " => " ")
```
tecosaur, line 469: What's with the space in this regexp?

src/trials.jl
vyu: Sure, I've changed it to skip the `histrow` assignment. (`enumerate(view(hist, r, :))` to avoid the allocation with array slicing.)

test/TrialsTests.jl
vyu: I'm not sure what you mean. This test is diagnostic against the current and the previous `remtrailingzeros` bugfixes by checking that "1.01 ns" is not printed as "1.010 ns" or "11 ns". It is also diagnostic against the bugfix for compability with Julia 1.0 by calling `show` on a `Trial`. It seems better to commit them together.

src/trials.jl
vyu: Oh, you might have looked at this before I posted my comment that explains why I changed this function. To recap: `timestr` includes units (e.g., `1.010 ns`), so we need to anchor to a space instead of the end of line. In any case, the spaces on each side act as a positive lookahead, so this is equivalent to `r"\.?0+(*pla: )" => ""`. Without them the function would revert to unconditionally removing the decimal point and right-adjacent zeros (which might not be trailing).

src/trials.jl
tecosaur: Cool.

src/trials.jl
tecosaur: Ah right, I just thought this was a random test.

test/TrialsTests.jl
tecosaur: Ah right, I just thought this was a random test.

src/trials.jl
tecosaur: I did look at this before reading your comment, this seems sensible.

test/TrialsTests.jl
```diff
@@ -233,4 +233,17 @@ else
...
+@test sprint(show, "text/plain", trial) == """
+BenchmarkTools.Trial: 2 samples with 1 evaluation.
+ Range (min … max):  1.000 ns … 1.010 ns  ┊ GC (min … max): 0.00% … 0.00%
+ Time  (median):     1.005 ns             ┊ GC (median):    0.00%
+ Time  (mean ± σ):   1.005 ns ± 0.007 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%
+
+  █                                                       █
```
vchuravy, line 243: @tecosaur one of the bars here is misaligned?

tecosaur (approved):

test/TrialsTests.jl
tecosaur: It looks fine to me.

image

Issue #214, opened 2021-05-25
Memory of interpolated variables is never freed


mfsch:
This might be the intended behavior, but I’ve noticed that when running multiple benchmarks in a loop, the data of the variables that are interpolated into the benchmark accumulates until the system eventually runs out of memory. A simplified demonstration of this behavior:

```julia
function bmk()
    a = zeros(128*1024*1024) # 1GB
    b = @benchmarkable $a .+= 1
    run(b)
end
```

Repeated calls of this function each increase the memory usage of the Julia process by 1GB. My actual usage where I ran into this issue was a bit more complex, where I wanted to measure the performance of some of my functions for different problem sizes, including sizes that take up a major part of the system’s memory.

If this is the intended behavior, would it be possible to document it in the section about interpolated variables? And are there any workarounds other than starting a new Julia process for each problem size?

vchuravy:
Duplicate of #127 

I think the general feeling is that this is something that needs fixing, 
but no one had the time/energy yet

PR #247, opened 2021-08-09
Add note about memory of interpolated variables to manual

mfsch:
As mentioned in https://github.com/JuliaCI/BenchmarkTools.jl/issues/214, I think it would be a good idea to document somewhere that variables interpolated into a benchmark won’t get garbage collected. Feel free to change or move the note.

### DIFF
--- a/docs/src/manual.md
+++ b/docs/src/manual.md
@@ -208,6 +208,9 @@ julia> let x = 1
        end
 ```
 
+!!! warning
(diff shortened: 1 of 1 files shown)


vchuravy:
Thanks! That's a good caveat to add.

mcabbott:
Is this also true of objects created in `setup=` block? I think so, in which case this might be worth also mentioning.

vchuravy:
Fixed by #265

PR #250, opened 2021-08-25
Make sure times are sorted in the `at-benchmark` histogram

### LINKED ISSUE #249: Confusing numbers in histogram
Trying to reproduce the [benchmarks in this discourse post](https://discourse.julialang.org/t/broadcasted-assignment-slower-than-for-loop-when-indices-are-arbitrary/66729), I've got
```
ulia> @benchmark $a[$w] .= $b[$w]
 Range (min … max):  10.176 μs …  1.150 ms  ┊ GC (min … max): 0.00% … 98.96%
 Time  (median):     10.987 μs              ┊ GC (median):    0.00%
 Time  (mean ± σ):   13.660 μs ± 17.326 μs  ┊ GC (mean ± σ):  3.08% ±  2.71%

  █                                                            
  █▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁ ▁
  53.4 μs         Histogram: frequency by time        11.1 μs <

 Memory estimate: 39.33 KiB, allocs estimate: 2.
```
<s>The mean time is `13.660 μs ± 17.326 μs`, I haven't looked at the code but the standard deviation larger than the average time itself looks suspiciously wrong (or maybe large one-sided outliers are screwing it up?).</s> :point_left:  Yes, I believe the problem is that the distribution isn't normal, there are large outliers only on one side, so the standard deviation is messed up.  The issue below still stands though :point_down: 

Also, note that the histogram has `53.4 μs` on the left and `11.1 μs` on the right: I have no idea how to read this histogram, is it reversed?

giordano:
Before (see #249):
```julia
julia> b = @benchmark $a[$w] .= $b[$w];

julia> b
 Range (min … max):   9.983 μs …  1.133 ms  ┊ GC (min … max): 0.00% … 98.94%
 Time  (median):     10.765 μs              ┊ GC (median):    0.00%
 Time  (mean ± σ):   13.441 μs ± 19.772 μs  ┊ GC (mean ± σ):  3.75% ±  2.72%

  █                                                            
  █▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁ ▁
  39.3 μs         Histogram: frequency by time        20.5 μs <

 Memory estimate: 38.89 KiB, allocs estimate: 2.
```
After:
```julia
julia> b
 Range (min … max):   9.983 μs …  1.133 ms  ┊ GC (min … max): 0.00% … 98.94%
 Time  (median):     10.765 μs              ┊ GC (median):    0.00%
 Time  (mean ± σ):   13.441 μs ± 19.772 μs  ┊ GC (mean ± σ):  3.75% ±  2.72%

  ▆▇█▅▄▂        ▁▁              ▂▂▂▃▃▂▃▃▂▁▁▁                  ▂
  ███████▅▆▆▅▄▄▄██▅▄▄▄▃▁▃▅▃▄▄▆▇████████████████▇█▆▇▇▆▅▆▆▆▅▅▅▆ █
  9.98 μs      Histogram: log(frequency) by time      29.2 μs <

 Memory estimate: 38.89 KiB, allocs estimate: 2.
```
I don't know how no one else noted that all histograms were broken.

I'm not an expert of `BenchmarkTools` internals, please triple check I didn't mess up anything.  For example, I'm assuming `times` and `gctimes` should be sorted together, but I'm not completely sure about it.

Fix #249.  CC: @tecosaur

### DIFF
--- a/src/trials.jl
+++ b/src/trials.jl
@@ -344,6 +344,10 @@ function Base.show(io::IO, ::MIME"text/plain", t::Trial)
     print(io, "BenchmarkTools.Trial: ", length(t), " sample", if length(t) > 1 "s" else "" end,
           " with ", t.params.evals, " evaluation", if t.params.evals > 1 "s" else "" end ,".\n")
 
+    perm = sortperm(t.times)
+    times = t.times[perm]
+    gctimes = t.gctimes[perm]
+
     if length(t) > 1
         med = median(t)
         avg = mean(t)
@@ -353,7 +357,7 @@ function Base.show(io::IO, ::MIME"text/plain", t::Trial)
 
         medtime, medgc = prettytime(time(med)), prettypercent(gcratio(med))
         avgtime, avggc = prettytime(time(avg)), prettypercent(gcratio(avg))
-        stdtime, stdgc = prettytime(time(std)), prettypercent(Statistics.std(t.gctimes ./ t.times))
+        stdtime, stdgc = prettytime(time(std)), prettypercent(Statistics.std(gctimes ./ times))
         mintime, mingc = prettytime(time(min)), prettypercent(gcratio(min))
         maxtime, maxgc = prettytime(time(max)), prettypercent(gcratio(max))
 
@@ -361,8 +365,8 @@ function Base.show(io::IO, ::MIME"text/plain", t::Trial)
         allocsstr = string(allocs(min))
     elseif length(t) == 1
         print(io, pad, " Single result which took ")
-        printstyled(io, prettytime(t.times[1]); color=:blue)
-        print(io, " (", prettypercent(t.gctimes[1]/t.times[1]), " GC) ")
+        printstyled(io, prettytime(times[1]); color=:blue)
+        print(io, " (", prettypercent(gctimes[1]/times[1]), " GC) ")
         print(io, "to evaluate,\n")
         print(io, pad, " with a memory estimate of ")
         printstyled(io, prettymemory(t.memory[1]); color=:yellow)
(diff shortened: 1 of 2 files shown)


src/trials.jl
```diff
@@ -344,6 +344,10 @@ function Base.show(io::IO, ::MIME"text/plain", t::Trial)
     print(io, "BenchmarkTools.Trial: ", length(t), " sample", if length(t) > 1 "s" else "" end,
           " with ", t.params.evals, " evaluation", if t.params.evals > 1 "s" else "" end ,".\n")
 
+    perm = sortperm(t.times)
+    times = t.times[perm]
+    gctimes = t.gctimes[perm]
```
tecosaur, line 349: I'm not sure that we actually want `times` and `gctimes` to be paired like this. When we say "max gc time" I'd expect that to be the maximum GC time of the runs personally.

vchuravy (approved):

src/trials.jl
giordano: So they should be sorted independently? Or gc times not sorted at all?

vchuravy: See https://github.com/JuliaCI/BenchmarkTools.jl/pull/250/files#diff-4950e391a6cdecb0a6ab326a117041a5b97d914db33c6234d7f312a69790258cR360 they need to be sorted the same so that the calculation there makes sense.

vchuravy: >  When we say "max gc time" I'd expect that to be the maximum GC time of the runs personally.

IIRC this has always been correlated e.g. we reported `gcratio(maximum(t))`.

tecosaur: Hmm, I see.

PR #255, opened 2021-09-16
Support manual range for hist display

timholy:
This allows one to manually set the range of the histogram during
display via an `IOContext`. The main use-case is for comparing
multiple `@benchmark` runs using a common axis for the
histogram.

Here's an example from the manual:

image

CC @tecosaur

### DIFF
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -37,6 +37,9 @@ julia> @btime sin(x) setup=(x=rand())
 0.49587200950472454
 ```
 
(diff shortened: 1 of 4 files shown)


tecosaur:
I'm very happy to see this PR, thanks Tim! I've been wanting to make the histogram customisable like this for some time now, but have just been too busy. To me, the way this has been implemented looks pretty reasonable. A `setup` parameter would likely be a bit easier to use … but this isn't "setup", and no better alternative comes to my mind.

vchuravy (approved):

vchuravy:
This is great!

Issue #237, opened 2021-07-16
log(frequency) in histograms confusing


karajan9:
While I was benchmarking two very similar functions I noticed the histograms looked very different, which surprised me. This was because the second run switched to a log display of the histogram.
![image](https://user-images.githubusercontent.com/6549509/125929852-a05911bd-8130-4e01-912e-40f1fa5618b3.png)
(I overwrote the function so they share the same name)

Having the option of a log histogram means I need to check what histogram I'm getting every time. I personally have also trouble reading a log histogram without thinking through what that actually means for my numbers. If one time dominates the histogram I would prefer to see it that way -- I assume the histogram is supposed to be a summary and a single dominant bar would be a good summary in that case. Therefore I would suggest always showing the histogram against the frequency, maybe with an option to display the log histogram.

If the behaviour should stay the same I would like to see an option to pick the histogram type so I can at least compare histograms of measurements easily.

tecosaur:
Hmmm, yea - this seems reasonable. I don't have a good idea right at this moment how to make it optional, but I'll get back to this at some point. In the meantime, the shape/profile you're getting isn't as extreme as the cases which I added a log-frequency histogram for, so the heuristic could maybe do with tweaking. The current logic is:
```
# if median size of (bins with >10% average data/bin) is less than 5% of max bin size, log the bin sizes
```

karajan9:
Sounds good.

If we are arguing a good default (of course log plots are fine in general, I just don't like them here as a part of the default) maybe gathering a few more voices would be a good idea to see if my opinion is more rule or more exception.

tecosaur:
I added them because in some circumstances a non-log plot can push all the other values to round to 0 on the bar graph, which gives the false impression that the executions all took that one time to execute. However, if as can happen that "one time" is say 2us, but 1/100th of the results were 1ms then that rare result which is visually buried actually can account for a large portion of the total time elapsed. This may sound overly contrived, but I have seen examples like that (though I cannot recall one the spot).

mkborregaard:
Sure, but would having normal axes not actually help in spotting this phenomenon, and making them automatically log hide it?

tecosaur:
I don' quite see what you mean. Normal axis hide this phenomenon, and log axis make it visible.

mkborregaard:
I guess that's hard to discuss without a sample distribution and a criterion for what diagnostic aspect of the benchmarking runs we are trying to isolate.

[...]

PR #257, opened 2021-09-18
Simplify control logic for :logbins

timholy:
There wasn't an actual bug, but the logic was so tortured that when I
looked back at it, I thought there was. This should be easier on the eyes.

It also adds some nice tests that may be useful templates for future improvements to the display (e.g., xref #237).


timholy:
Since this has no functional consequence and adds tests (never a bad thing), I'll merge this (it's not really worth review time).

PR #258, opened 2021-09-18
Show mean time in `@btime`

mcabbott:
There is room to show a bit more information with `@btime`, and I think the next number to show after the minimum is the mean. So this PR does this:

```julia
julia> @btime sum(log, rand(1000));
  min 4.441 μs, mean 5.325 μs (1 allocation: 7.94 KiB)

julia> @btime sum(log, $(rand(1000)));
  min 3.693 μs, mean 3.721 μs (0 allocations)
```
Before:
```
julia> @btime sum(log, $(rand(1000)));
  5.285 μs (0 allocations: 0 bytes)
```
Notice that it also doesn't bother writing "0 bytes" if there are no allocations, in the name of reducing clutter. I think that's safe but if anyone has a counter-example speak up.

It's possible that it should print something about GC time, but what number or numbers exactly would be best? Often the fastest run will not trigger this, but should it be the mean GC time (as a percentage), or the mean GC percentage, or something else?

Xref #217 for ongoing discussion about `@benchmark`'s histograms.

Needs docstring updates, and probably tests.


MasonProtter:
I think it’s useful to say `0 bytes` if there’s not allocations. One thing I really like about this format is that most users don’t realize when they use `@btime` that it’s reporting a minimum. Most assume it’s an average or median time.

mcabbott:
It could also say `min 1 allocation, 7.94 KiB` since that's what it is, but maybe it's usually stable enough for this not to matter.

src/execution.jl
```diff
@@ -550,28 +550,51 @@ this executes an expression, printing the time
...
         local $trial, $result = $BenchmarkTools.run_result($bench)
         local $trialmin = $BenchmarkTools.minimum($trial)
+        local $trialmean = $BenchmarkTools.mean($trial)
         local $trialallocs = $BenchmarkTools.allocs($trialmin)
-        println("  ",
-                $BenchmarkTools.prettytime($BenchmarkTools.time($trialmin)),
-                " (", $trialallocs , " allocation",
-                $trialallocs == 1 ? "" : "s", ": ",
-                $BenchmarkTools.prettymemory($BenchmarkTools.memory($trialmin)), ")")
+        $print("  min ", $BenchmarkTools.prettytime($BenchmarkTools.time($trialmin)),
```
mcabbott, line 587: Is there a reason to interpolate `$BenchmarkTools.prettymemory` not just `$prettymemory`? I see I left some and later dropped them.

src/execution.jl
```diff
@@ -550,28 +550,51 @@ this executes an expression, printing the time
...
+
+julia> @btime rand(1000);
+  min 724.432 ns, mean 1.462 μs (1 allocation, 7.94 KiB. GC mean 352 ns, 24.11%)
+```
 """
 macro btime(args...)
     _, params = prunekwargs(args...)
     bench, trial, result = gensym(), gensym(), gensym()
-    trialmin, trialallocs = gensym(), gensym()
+    trialmin, trialmean, trialallocs = gensym(), gensym(), gensym()
```
mcabbott, line 577: Is there a reason not to write `@gensym trialmin trialallocs` etc. here?

KristofferC:
> First they came for `@benchmark`...

I think there is great value in having `@btime` only show the number which is in most cases useful which is the minimum. For more data, there is `benchmark`.

mcabbott:
This is a fair point. And there's something to be said for matching `@time` (although that now adds compilation time, too). I'd also like to cut down `@benchmark` a bit, to make the step up smaller than it is now.

Can I persuade anyone to try out this branch in real life for a bit, to see if they like or hate it in actual use?

KristofferC:
The issue is that for cases where you would use `@btime` over `@benchmark` the mean is a worse statistic than the mean. If you want to show extra stuff, perhaps the mode could be shown, but the more you measure the more the mode will go towards the minimum anyway.

mcabbott:
Isn't the mean what's relevant for throughput? I guess this PR takes the position that it's the most useful *second* number. For example, with the allocating `sum(x .* y)` above, the GC time seems to be an under-estimate of the penalty: 16%, while almost 60% of the mean time is absent from the minimum. This seems relevant for micro-benchmarking a function you hope to use in a loop.

The second number `@time` prints is GC percentage, from just one run of course. I would argue that showing the percentage of the mean without showing the mean is weird, and usually GC does not run at all on the `min` run. And that the mean time is more useful anyway -- it's easier to have all numbers referring to the total time. If we want to keep the length shorter, deleting this would help.

The mode seems tricky, on what are essentially real numbers. Should they be binned first, or can we trust the clock to do that well enough? 
```
julia> b2 = @benchmark sum(log.(rand(10^4)));
    
julia> length(b2.times)
10000

julia> diff(sort(unique(b2.times)))
766-element Vector{Float64}:
     41.0
     84.0
     41.0
      1.0
     41.0
     42.0
     41.0
      1.0
     41.0
      1.0
      ⋮
```
This and the median might be contenders for the best *one* number. But, as you say, these will often approach the minimum... which `@btime` already prints, and is the easiest to explain, so this PR doesn't propose to change.

mcabbott:
Nice, somehow I never looked at the paper. 

I guess garbage collection is the major thing I had in mind, since it's very common to have different implementations with different allocations. The reported GC percentage is quite a large underestimate of how much time this is going to cost. This is a rare event, so something like mean(5-to-95th) will miss it. 

Agree this measure will be noisier than `min`, but the claim here is only that it's a sufficiently useful *second number* to be worth printing. If it's huge, that's a warning to investigate more closely. 

Is there another number which might better serve this purpose?

KristofferC:
> And if I include these functions as part of some larger long-running loop, then it seems the mean doesn't do badly as a prediction of how long the overall calculation will take, unlike the minimum:

It is because you are not running a representative benchmark. You are doing the equivalent of

```
 @btime sin.(rand(10^4:10^5))
```

and then saying that the minimum is a bad statistic for giving the mean value when running that code multiple times. 

Mean is bad. Every single noise of your computer, every latency based on the OS scheduler, etc contributes to the mean. Don't take it from me, take it from this dude https://youtu.be/vrfYLlR8X8k?t=1487 (the whole talk is worth looking at).

Again, make a `@bmean` or something, leave `@btime` alone.

chriselrod:
I want this.

I'm perfectly fine with naming it `@bmean` or something else to preserve `@btime`.

KristofferC:
Yes, call it something else.

chriselrod:
> Yes, call it something else.

Are you happy to rename it? Seems easy enough to make everyone happy this way.

gdalle:
@mcabbott doing some cleanup before a 2.0 release, would you wanna rename this `@bmean` and try to get it merged?

Issue #127, opened 2018-11-29
Pass setup/interpolated variables through as `run` arguments instead of closing over them in the harness


jrevels:
cc @maleadt 
cc @KristofferC (this approach could help with the optimization-brittleness issues you've said run into)

See my comment here: https://github.com/JuliaGPU/CuArrays.jl/issues/210#issuecomment-442860840

> Right now, interpolated variables get closed over in the benchmarking harness [here](https://github.com/JuliaCI/BenchmarkTools.jl/blob/f0f22b63d7076096141d0cc2e294246a3a38feb4/src/execution.jl#L299).
> 
> We should probably change this so that `BenchmarkTools.run` accepts a (splatted?) tuple of arguments that gets forwarded into the harness/kernel and bound to the interpolated variables. This way arguments can be passed in from top-level, which would likely yield better results than the current implementation (less brittle w.r.t. compiler optimizations) + make the implementation nicer (we could have a single harness definition instead of generating benchmark-specific ones, and probably factor out some other stuff as well).
> For `@btime`/`@benchmark`, this would not be a breaking change. It would probably need to be breaking for `@benchmarkable`, though - that would have to now e.g. return the Benchmark + the tuple/NamedTuple for its `run` arguments.

I'm not sure this would've been possible to implement easily in the pre-NamedTuple era, but now that we have them, we might be able to use them to forward variables through to the benchmark kernel. 

E.g. the whole `setup` expression could be rendered into a thunk that gets called with the interpolated variable arguments, and returns a NamedTuple of all the resulting setup variables.

KristofferC:
Yeah, I noticed this when reading https://discourse.julialang.org/t/when-does-the-garbage-collector-runs/15142/3 (which made me make https://github.com/JuliaCI/BenchmarkTools.jl/pull/117 but that, of course, doesn't change that values get captured in the generated function).

Is there any way to use anonymous functions which gets garbage collected and then release the resources they close over.

willow-ahrens:
This was fixed in https://github.com/JuliaCI/BenchmarkTools.jl/pull/265

PR #265, opened 2021-11-24
Ensure interpolated variables are garbage-collectable

ettersi:
With this PR, the values interpolated into an `@benchmarkable` are stored in the `Benchmark` object rather than interpolated directly into the benchmark code. This ensures that interpolated values can be garbage-collected once the `Benchmark` object goes out of scope, in contrast to the current behaviour where values interpolated into a benchmark expression will stick around forever. 

This PR would solve https://github.com/JuliaCI/BenchmarkTools.jl/issues/127. 

MWE demonstrating what works with this PR but fails otherwise:
```julia
# Assemble a benchmark with an interpolated value
x = []
x_finalized = false
finalizer(x->(global x_finalized=true), x)
b = @benchmarkable $x

# Free the memory
b = x = nothing
GC.gc()

# Check that the memory has indeed been freed
@test x_finalized
```

### DIFF
--- a/src/execution.jl
+++ b/src/execution.jl
@@ -9,6 +9,7 @@ gcscrub() = (GC.gc(); GC.gc(); GC.gc(); GC.gc())
 
 mutable struct Benchmark
     samplefunc
+    quote_vals
     params::Parameters
 end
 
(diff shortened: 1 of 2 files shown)


ettersi:
Bump.

vchuravy (approved):

mcabbott:
This also seems to have effects on constant propagation. Taking the example from the manual, before:
```
julia> a = 1; b = 2
2

julia> @btime $a + $b
  0.001 ns (0 allocations: 0 bytes)
3

julia> @btime $(Ref(a))[] + $(Ref(b))[]
  2.125 ns (0 allocations: 0 bytes)  # not 1.500 ns just because of noise I think
3

(@v1.5) pkg> st BenchmarkTools
Status `~/.julia/environments/v1.5/Project.toml`
  [6e4b80f9] BenchmarkTools v1.2.0
```
After:
```
julia> a = 1; b = 2
2

julia> @btime $a + $b
  1.500 ns (0 allocations: 0 bytes)
3

julia> @btime $(Ref(a))[] + $(Ref(b))[]
  1.500 ns (0 allocations: 0 bytes)
3

julia> @btime 1+2
  0.001 ns (0 allocations: 0 bytes)   # unchanged
3

julia> @btime $1 + $2
  2.083 ns (0 allocations: 0 bytes)
3

julia> @btime $1 + $2
  1.458 ns (0 allocations: 0 bytes)
3

(@v1.5) pkg> st BenchmarkTools
Status `~/.julia/environments/v1.5/Project.toml`
  [6e4b80f9] BenchmarkTools v1.2.1
```
I think that's better. Are there any examples left where an interpolated variable will defeat the benchmark? If not, perhaps the manual should be updated to remove the `Ref` trick in favour of `$`.

PR #275, opened 2022-02-04
Make use of Base.donotdelete if available

Keno:
Goes with https://github.com/JuliaLang/julia/pull/44036.

### DIFF
--- a/src/execution.jl
+++ b/src/execution.jl
@@ -479,6 +479,9 @@ function generate_benchmark_definition(eval_module, out_vars, setup_vars, quote_
         invocation = :($(Expr(:tuple, out_vars...)) = $(signature))
         core_body = :($(core); $(returns))
     end
+    if isdefined(Base, :donotdelete)
+        invocation = :(Base.donotdelete($invocation))
+    end
     return Core.eval(eval_module, quote
         @noinline $(signature_def) = begin $(core_body) end
         @noinline function $(samplefunc)($(Expr(:tuple, quote_vars...)), __params::$BenchmarkTools.Parameters)


vchuravy:
I am not sure that running this unconditionally is the best idea. One of the stated intentions for BenchmarkTools has always been that it will execute the JuliaCode as is, to match the actual behavior as close as possible. 

I see the utility for BaseBenchmarks, but maybe it should change the definition to include `dcebarrier(sin(x))`?

Keno:
I would think that most benchmarks don't actually have any side-effects, so as the compiler gets better I think we'll be deleting them more and more. I think people would expect that the return value from their benchmarks would get preserved (as it would if you just run it at the REPL, because then it gets assigned to `ans`).

KristofferC:
> One of the stated intentions for BenchmarkTools has always been that it will execute the JuliaCode as is, to match the actual behavior as close as possible

A better intention seems to me to have BenchmarkTools be as useful as possible. And when it optimizes away the whole benchmark, it is not being useful. The hoops you have to jump through now with interpolating `Ref` etc just to get a useful answer is annoying. This seems like an improvement.

Keno:
Base is ready to merge, so let's get this in, so we can do the Base nanosoldier run.

vchuravy (approved):

giordano:
After this PR `@btime` doesn't return the output of the operation anymore:
```julia
julia> using BenchmarkTools; @btime 1 + 2
  1.413 ns (0 allocations: 0 bytes)

julia>
```

giordano:
You mean like this:
```patch
--- a/src/execution.jl
+++ b/src/execution.jl
@@ -480,7 +480,10 @@ function generate_benchmark_definition(eval_module, out_vars, setup_vars, quote_
         core_body = :($(core); $(returns))
     end
     if isdefined(Base, :donotdelete)
-        invocation = :(Base.donotdelete($invocation))
+        invocation = :(let x = $invocation
+                           Base.donotdelete(x)
+                           x
+                       end)
     end
     return Core.eval(eval_module, quote
         @noinline $(signature_def) = begin $(core_body) end
```
? I'm not entirely sure whether `x` needs to be inside `donotdelete` or not (I think so because that should be the point of this function).

Keno:
Yes, that patch would work

Issue #221, opened 2021-06-14
@benchmarkset seems to be missing cases


awadell1:
When I run the example code from the docs I get the following

```julia
julia> @benchmarkset "suite" for k in 1:5
           @case "case $k" rand($k, $k)
       end
1-element BenchmarkTools.BenchmarkGroup:
  tags: []
  "suite" => 1-element BenchmarkTools.BenchmarkGroup:
	  tags: []
	  "case 5" => Benchmark(evals=1, seconds=5.0, samples=10000)

```

Instead of what I would expect:

```
5-element BenchmarkTools.BenchmarkGroup:
  tags: []
  "suite" => 1-element BenchmarkTools.BenchmarkGroup:
	  tags: []
	  "case 1" => Benchmark(evals=1, seconds=5.0, samples=10000)
	  "case 2" => Benchmark(evals=1, seconds=5.0, samples=10000)
	  "case 3" => Benchmark(evals=1, seconds=5.0, samples=10000)
	  "case 4" => Benchmark(evals=1, seconds=5.0, samples=10000)
	  "case 5" => Benchmark(evals=1, seconds=5.0, samples=10000)
```

Am I missing something? @Roger-luo

LilithHafner:
Same issue here.

PR #284, opened 2022-06-06
Make @benchmarkset work with for loop

skleinbo:
Closes https://github.com/JuliaCI/BenchmarkTools.jl/issues/221

Fixes a bug where every iteration of the for loop
would overwrite the previous benchmark case.

Test.

### DIFF
--- a/src/groups.jl
+++ b/src/groups.jl
@@ -391,7 +391,7 @@ function benchmarkset_block(title, ex::Expr)
     stack = GlobalRef(BenchmarkTools, :benchmark_stack)
     quote
         let $(Symbol("#root#")) = last($stack)
-            $(Symbol("#root#"))[$title] = $(Symbol("#suite#")) = BenchmarkGroup()
+            $(Symbol("#suite#")) = $(Symbol("#root#"))[$title]
             push!($stack, $(Symbol("#suite#")))
             $ex
             pop!($stack)
--- a/test/GroupsTests.jl
+++ b/test/GroupsTests.jl
@@ -293,6 +293,16 @@ g2[[1, "a", :b]] = "hello"  # should create higher levels on the fly
     @test haskey(g1["test set"], "test case 1")
     @test haskey(g1["test set"], "test case 2")
 end
+
+@testset "benchmarkset for loop" begin
+  g1 = @benchmarkset "test set" for k in 1:2
+     @case "test case $k" $k + $k
+  end
+
+  @test haskey(g1, "test set")
+  @test haskey(g1["test set"], "test case 1")
+  @test haskey(g1["test set"], "test case 2")
+end
 # pretty printing #
 #-----------------#
 


src/groups.jl
```diff
@@ -364,7 +364,11 @@ function benchmarkset_block(title, ex::Expr)
     stack = GlobalRef(BenchmarkTools, :benchmark_stack)
     quote
         let $(Symbol("#root#")) = last($stack)
-            $(Symbol("#root#"))[$title] = $(Symbol("#suite#")) = BenchmarkGroup()
+            if !haskey($(Symbol("#root#")), $title)
+                $(Symbol("#root#"))[$title] = $(Symbol("#suite#")) = BenchmarkGroup()
+            else
+                $(Symbol("#suite#")) = $(Symbol("#root#"))[$title]
```
gdalle, line 370: Can you explain to me why `$(Symbol("#suite#"))` is set based on `$(Symbol("#root#"))[$title]` here, but two lines above it's the reverse? I'm just discovering this code base

skleinbo: `#suite#` should refer to the BenchmarkGroup `$title` under `#root#`. However, that group might not exist (yet) and so we check if it does, and if not initiate it.

The problem with the original code was, that it only had
```
$(Symbol("#root#"))[$title] = $(Symbol("#suite#")) = BenchmarkGroup()
```
and thus the BenchmarkGroup `$title` was overwritten at every loop iteration.


Here is the generated code from the test case (formatted with `prettify` from `MacroTools`)
```
quote
    if isempty(BenchmarkTools.benchmark_stack)
        push!(BenchmarkTools.benchmark_stack, (BenchmarkGroup)())
    end
    for k = 1:2
        let goat = last(BenchmarkTools.benchmark_stack)
            if !(haskey(goat, "test set"))
                goat["test set"] = (hummingbird = BenchmarkGroup())
            else
                hummingbird = goat["test set"]
            end
            push!(BenchmarkTools.benchmark_stack, hummingbird)
            hummingbird["test case $(k)"] = BenchmarkTools.generate_benchmark_definition(Main, Symbol[], Any[], [Symbol("##k#316"), Symbol("##k#317")], (k, k), $(Expr(:copyast, :($(QuoteNode(:(var"##k#316" + var"##k#317")))))), $(Expr(:copyast, :($(QuoteNode(nothing))))), $(Expr(:copyast, :($(QuoteNode(nothing))))), BenchmarkTools.Parameters())
            pop!(BenchmarkTools.benchmark_stack)
        end
    end
    if length(BenchmarkTools.benchmark_stack) == 1
        pop!(BenchmarkTools.benchmark_stack)
    end
end
```

gdalle: Does this get easier due to the recently-merged #309 ?

skleinbo: It probably does! I hadn't seen that PR. I'll check it out.

gdalle: Yeah, I just got ~~hired~~ trapped as a maintainer so I'm cleaning up the house ^^

skleinbo: Thanks for taking on the responsibility!

I've adapted the code according to #309. Very convenient.

gdalle: OK I don't fully understand what's going on but the test seems to work and the code change is now one line so I'm choosing trust

PR #290, opened 2022-10-05
Add Example Code to Common Workflow

### LINKED ISSUE #289: How do you save benchmark results?
https://juliaci.github.io/BenchmarkTools.jl/dev/manual/#Caching-Parameters

1. Start a Julia session
2. Execute a benchmark suite using an old version of your package
3. Save the results somehow (e.g. in a JSON file)

how? The following description is about saving the parameters to a file, but how do I actually save the Benchmark Results?

FelixBenning:
fix #289

### DIFF
--- a/docs/src/manual.md
+++ b/docs/src/manual.md
@@ -857,10 +857,23 @@ A common workflow used in BenchmarkTools is the following:
 
 1. Start a Julia session
 2. Execute a benchmark suite using an old version of your package
-3. Save the results somehow (e.g. in a JSON file)
+    ```julia
+    results = run(suite, verbose = true)
+    ```
+4. Save the results somehow (e.g. in a JSON file)
+    ```julia
+    BenchmarkTools.save("results.json", median(results))
+    ```
 4. Start a new Julia session
 5. Execute a benchmark suite using a new version of your package
-6. Compare the new results with the results saved in step 3 to determine regression status
+   ```julia
+   results = run(suite, verbose = true)
+   ```
+7. Compare the new results with the results saved in step 3 to determine regression status
+    ```julia
+    old_results = BenchmarkTools.load("results.json")
+    BenchmarkTools.judge(median(results), old_results)
+    ```
 
 There are a couple of problems with this workflow, and all of which revolve around parameter tuning (which would occur during steps 2 and 5):
 


docs/src/manual.md
```diff
@@ -857,10 +857,23 @@ A common workflow used in BenchmarkTools is the following:
 
 1. Start a Julia session
 2. Execute a benchmark suite using an old version of your package
-3. Save the results somehow (e.g. in a JSON file)
+    ```julia
+    results = run(suite, verbose = true)
```
gdalle, line 861:
```suggestion
    old_results = run(suite, verbose = true)
```

docs/src/manual.md
```diff
@@ -857,10 +857,23 @@ A common workflow used in BenchmarkTools is the following:
...
+    ```
+4. Save the results somehow (e.g. in a JSON file)
+    ```julia
+    BenchmarkTools.save("results.json", median(results))
```
gdalle, line 865:
```suggestion
    BenchmarkTools.save("old_results.json", median(old_results))
```

gdalle: Why the median instead of the minimum?

docs/src/manual.md
```diff
@@ -857,10 +857,23 @@ A common workflow used in BenchmarkTools is the following:
...
+    ```
 4. Start a new Julia session
 5. Execute a benchmark suite using a new version of your package
-6. Compare the new results with the results saved in step 3 to determine regression status
+   ```julia
+   results = run(suite, verbose = true)
+   ```
+7. Compare the new results with the results saved in step 3 to determine regression status
+    ```julia
+    old_results = BenchmarkTools.load("results.json")
```
gdalle, line 874:
```suggestion
    old_results = BenchmarkTools.load("old_results.json")
```

docs/src/manual.md
```diff
@@ -857,10 +857,23 @@ A common workflow used in BenchmarkTools is the following:
...
+    BenchmarkTools.judge(median(results), old_results)
```
gdalle, line 875: For consistency, it would be nice if `results` and `old_results` were comparable (either both aggregated or none of them)

gdalle (changes requested):

gdalle:
Could you also add docstrings to `BenchmarkTools.save` and `BenchmarkTools.load`, and make sure they show up in the docs? That would be an amazing addition

docs/src/manual.md
FelixBenning: I don't know - there might be things where you can get lucky? Cache hit type of things. Or maybe you try to figure out if something is prime and your algorithm just tries to divide it by all numbers 2, 3, ... and it is lucky because it is divisible by two but that is not representative of the typical performance.

Median performance just seemed reasonable to me

docs/src/manual.md
gdalle: going for the minimum since that is the default quantity in BenchmarkTools

gdalle: tweaking a little

docs/src/manual.md
gdalle: minimum is the default quantity in BenchmarkTools (the one showed by `@btime`) so I'm going with that one

gdalle (approved):

PR #292, opened 2022-10-28
Add `@ballocations` macro for getting number of allocations

nickrobinson251:
- matches to the `@allocations` macro recently added to Julia Base (https://github.com/JuliaLang/julia/pull/47367)

### DIFF
--- a/src/execution.jl
+++ b/src/execution.jl
@@ -557,7 +557,7 @@ end
 """
     @ballocations expression [other parameters...]
 
-Similar to the `@allocs` macro included with Julia (v1.9+),
+Similar to the `@allocations` macro included with Julia (v1.9+),
 this returns the *number of allocations* when executing
 a given expression. It uses the `@benchmark` macro, however,
 and accepts all of the same additional parameters as `@benchmark`.
--- a/src/execution.jl
+++ b/src/execution.jl
@@ -539,16 +539,36 @@ end
     @ballocated expression [other parameters...]
 
 Similar to the `@allocated` macro included with Julia,
-this returns the number of bytes allocated when executing
+this returns the *number of bytes allocated* when executing
 a given expression.   It uses the `@benchmark`
 macro, however, and accepts all of the same additional
 parameters as `@benchmark`.  The returned allocations
 correspond to the trial with the *minimum* elapsed time measured
 during the benchmark.
+
+See also `@ballocs`.
 """
 macro ballocated(args...)
     return esc(quote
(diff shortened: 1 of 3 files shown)


nickrobinson251:
bump :) is someone available to review this please?

KristofferC:
I don't really understand the purpose of this macro and the `@ballocated` one. Why would you run a function potentially thousands of times (until the execution time is within some noise level) to count the number of allocations (which is not noisy). Isn't one for compiling the function and one for counting the allocations enough?

nickrobinson251:
i hope to use it the same way `@ballocated` is already used in e.g. the NamedDims tests: https://github.com/invenia/NamedDims.jl/search?q=ballocated

running these functions many times is no issue (since they're written to be basically no cost), but it's a lot nicer (i.e. less burden on the package developers writing the tests) to use `@ballocated` rather than having to makes sure things have already compiled and then using `@allocated`

gdalle:
> but i think setting default evals/samples could make sense -- should i make that change?

Sounds like a good idea to me, although it will have to be thoroughly documented for users not to be surprised.

src/execution.jl
```diff
@@ -539,19 +539,39 @@ end
...
 macro ballocated(args...)
     return esc(quote
         $BenchmarkTools.memory($BenchmarkTools.minimum($BenchmarkTools.@benchmark $(args...)))
     end)
 end
 
+"""
+    @ballocations expression [other parameters...]
+
+Similar to the `@allocs` macro included with Julia (v1.9+),
```
gdalle, line 560:
```suggestion
Similar to the `@allocations` macro included with Julia (v1.9+),
```

gdalle, line 560:
```suggestion
Similar to the `@allocations` macro included with Julia (v1.9+),
```

src/execution.jl
```diff
@@ -539,19 +539,39 @@ end
...
+this returns the *number of allocations* when executing
+a given expression. It uses the `@benchmark` macro, however,
+and accepts all of the same additional parameters as `@benchmark`.
+The returned allocations correspond to the trial with
+the *minimum* elapsed time measured during the benchmark.
+
```
gdalle, line 566: Can we set a default value `evals=1 samples=2` for this one? And explain it in the docstring?

gdalle (changes requested):

src/execution.jl
gdalle: Then again it would be incoherent with the other `@bstuff` macros so I'm torn

nickrobinson251:
superseded by https://github.com/JuliaCI/BenchmarkTools.jl/pull/383

PR #309, opened 2023-05-09
Automatically create nested `BenchmarkGroup` on access

### LINKED ISSUE #308: Automatically create keys in BenchmarkGroup
Similar to how [EasyConfig.jl](https://github.com/JuliaComputing/EasyConfig.jl) works, it would be great if I could automatically create keys with `setindex!`, like:

```julia
suite = BenchmarkGroup()

suite["f1"]["param1"]["param2"] = @benchmarkable f1($param1, $param2)
# ^ Automatically creates nested BenchmarkGroup()
```

rather than having to do `suite["f1"] = BenchmarkGroup()` and so on.

MilesCranmer:
Fixes #308

This lets you perform, e.g.,

```julia
suite = BenchmarkGroup()

for T in [Float32, Float64], n in [10, 100]
    suite["sum"][T][n] = @benchmarkable sum(x) setup=(x=randn($T, $n))
end
```

without needing to manually create each benchmark group. Normally you would have to do something like:

```
suite["sum"] = BenchmarkGroup()

for T in [Float32, Float64], n in [10, 100]
    n == 10 && (suite["sum"][T] = BenchmarkGroup())
    suite["sum"][T][n] = @benchmarkable sum(x) setup=(x=randn($T, $n))
end
```

This is similar to how [EasyConfig.jl](https://github.com/JuliaComputing/EasyConfig.jl) works, in that `getindex` will automatically create each required `BenchmarkGroup()`.

The downside is that accessing the wrong key will create empty groups. Thus, there is the `clear_empty!` command to clean these up (I added docs too).

### DIFF
--- a/docs/src/manual.md
+++ b/docs/src/manual.md
@@ -529,6 +529,19 @@ julia> suite
 As you might imagine, `BenchmarkGroup` supports a subset of Julia's `Associative` interface. A full list of
 these supported functions can be found [in the reference document](reference.md#benchmarkgrouptagsvector-datadict).
 
+One can also create a nested `BenchmarkGroup` simply by passing the keys, similar to how
+`mkdir -p` works:
+
+```julia
+suite2 = BenchmarkGroup()
+
+suite2["my"]["nested"]["benchmark"] = @benchmarkable sum(randn(32))
+```
+
+which will result in a hierarchical benchmark without us needing to create each `BenchmarkGroup` ourselves.
+(Note that keys are automatically created upon access, even if a key does not exist. Thus, if you wish
+to empty the unused keys, you can use `clear_empty!` to do so)
+
 ### Tuning and running a `BenchmarkGroup`
 
 Similarly to individual benchmarks, you can `tune!` and `run` whole `BenchmarkGroup` instances (following from the previous section):
--- a/src/groups.jl
+++ b/src/groups.jl
@@ -23,16 +23,28 @@ function addgroup!(suite::BenchmarkGroup, id, args...)
     return g
 end
 
+clear_empty!(x) = x
+function clear_empty!(group::BenchmarkGroup)
+    for (k,v) in pairs(group)
+        if v isa BenchmarkGroup && isempty(v)
+            delete!(group, k)
+        else
+            clear_empty!(v)
+        end
+    end
+    group
+end
+
 # Dict-like methods #
 #-------------------#
 
 Base.:(==)(a::BenchmarkGroup, b::BenchmarkGroup) = a.tags == b.tags && a.data == b.data
 Base.copy(group::BenchmarkGroup) = BenchmarkGroup(copy(group.tags), copy(group.data))
 Base.similar(group::BenchmarkGroup) = BenchmarkGroup(copy(group.tags), empty(group.data))
-Base.isempty(group::BenchmarkGroup) = isempty(group.data)
+Base.isempty(group::BenchmarkGroup) = isempty(clear_empty!(group).data)
 Base.length(group::BenchmarkGroup) = length(group.data)
-Base.getindex(group::BenchmarkGroup, k) = getindex(group.data, makekey(k))
-Base.getindex(group::BenchmarkGroup, k...) = getindex(group.data, makekey(k))
+Base.getindex(group::BenchmarkGroup, k) = get!(group.data, makekey(k), BenchmarkGroup())
+Base.getindex(group::BenchmarkGroup, k...) = get!(group.data, makekey(k), BenchmarkGroup())
 Base.setindex!(group::BenchmarkGroup, v, k) = setindex!(group.data, v, makekey(k))
 Base.setindex!(group::BenchmarkGroup, v, k...) = setindex!(group.data, v, makekey(k))
 Base.delete!(group::BenchmarkGroup, k) = delete!(group.data, makekey(k))
--- a/src/BenchmarkTools.jl
+++ b/src/BenchmarkTools.jl
@@ -55,7 +55,8 @@ export BenchmarkGroup,
        addgroup!,
        leaves,
        @benchmarkset,
-       @case
+       @case,
+       clear_empty!
 
 ######################
 # Execution Strategy #


MilesCranmer:
@vchuravy @ranocha would one of you be up for reviewing this? Note that this is backwards compatible so won’t cause issues in existing code. It simply makes it significantly easier to set up hierarchical benchmark suites over different param combinations.

ranocha (commented):
I'm not a developer/maintainer of BenchmarkTools.jl and can't really judge this. All I can say is that you should also add some tests.

MilesCranmer:
Thanks. Will definitely add tests; just wanted to have this PR “fail fast” if this feature is a nonstarter

ranocha:
> if this feature is a nonstarter

I'm not a developer/maintainer, so I can't comment on this

MilesCranmer:
@vchuravy would you be able to take a look at this? I can add tests if this looks okay as a feature.

gdalle:
@vchuravy just made me a maintainer (hurray 🤣) and it turns out I ran into this very problem last week, so I'm gonna say the feature is a welcome one. Please add tests and I'll review!

MilesCranmer:
Done!

gdalle:
I need to fix CI cause #305 broke it for unsupported Julia versions, then we'll re-run the tests

gdalle:
OK tests are failing outside of 1.5 and 1.7, that's not on me :)

MilesCranmer:
Oops, fixed!

docs/src/manual.md
```diff
@@ -529,6 +529,19 @@ julia> suite
 As you might imagine, `BenchmarkGroup` supports a subset of Julia's `Associative` interface. A full list of
 these supported functions can be found [in the reference document](reference.md#benchmarkgrouptagsvector-datadict).
 
+One can also create a nested `BenchmarkGroup` simply by passing the keys, similar to how
+`mkdir -p` works:
```
gdalle, line 533: It's a bit niche, not sure that precision helps a lot of users. For instance I know what `mkdir` does but I've never heard about the `-p` flag.

docs/src/manual.md
```diff
@@ -529,6 +529,19 @@ julia> suite
...
+
+```julia
+suite2 = BenchmarkGroup()
+
+suite2["my"]["nested"]["benchmark"] = @benchmarkable sum(randn(32))
+```
+
+which will result in a hierarchical benchmark without us needing to create each `BenchmarkGroup` ourselves.
+(Note that keys are automatically created upon access, even if a key does not exist. Thus, if you wish
+to empty the unused keys, you can use `clear_empty!` to do so)
```
gdalle, line 543: `clear_empty!(suite)` or `clear_empty!(suite, key1, key2, ...)`?

src/groups.jl
```diff
@@ -23,16 +23,26 @@ function addgroup!(suite::BenchmarkGroup, id, args...)
...
+end
+
 # Dict-like methods #
 #-------------------#
 
 Base.:(==)(a::BenchmarkGroup, b::BenchmarkGroup) = a.tags == b.tags && a.data == b.data
 Base.copy(group::BenchmarkGroup) = BenchmarkGroup(copy(group.tags), copy(group.data))
 Base.similar(group::BenchmarkGroup) = BenchmarkGroup(copy(group.tags), empty(group.data))
-Base.isempty(group::BenchmarkGroup) = isempty(group.data)
+Base.isempty(group::BenchmarkGroup) = isempty(clear_empty!(group).data)
```
gdalle, line 42: I don't like the fact that this is implicitly mutating

docs/src/manual.md
MilesCranmer: Ah, good point. Seems like I’ve spent so much time in a unix terminal that I managed to convince myself this was common knowledge 😂

docs/src/manual.md
MilesCranmer: `clear_empty!` finds the keys for you (recursively). You would prefer it to be manually used?

gdalle: No the recursive behavior is needed, I was just suggesting you make it explicit :)

MilesCranmer: Oh I see, sounds good!

src/groups.jl
MilesCranmer: I think because subgroups are now created automatically upon access, pruning them eagerly like this doesn’t feel so much like a “mutation” (though it technically is), but more like a reduction to a standard form. ie

```julia
suite[1] = BenchmarkGroup()
isempty(suite[1])
suite[1][2] = ...
```

which works the exact same as before, due to this automatic creation.

This is used as a part of `clear_empty!` — you first need to clear out all empty children before seeing whether a parent `BenchmarkGroup()` is also empty. (Borrowed this approach from EasyConfig.jl.) This could be without changing the BenchmarkGroup(), but it would result in a recursive blow-up, because every time you call `isempty` inside `clear_empty!`, it would be creating copies of the BenchmarkGroup — rather than just working on a single copy and cleaning it up as it goes.

gdalle: Ok I understand! Maybe it would make sense to document it in a docstring, along with other key functions from Base, and add said docstrings to the API reference page?

docs/src/manual.md
MilesCranmer: Done!

docs/src/manual.md
MilesCranmer: Fixed

src/groups.jl
MilesCranmer: Done

gdalle (approved):

gdalle:
@MilesCranmer do you think this is a breaking change or a new feature?

MilesCranmer:
I think it’s a breaking change, so worth bumping the version number

gdalle:
OK then I'll wait until we fix a few longstanding bugs

MilesCranmer:
@gdalle I noticed a new version has not been released yet and I was wondering if you might be able to do so? Would be great to use this feature!

gdalle:
I have been hesitant cause it's a central package and breaking changes have been made, but a recent discussion on slack convinced me to move forward. Waiting on one PR then I'll do a bit of cleanup and tag 2.0

PR #327, opened 2023-08-01
Apply JuliaFormatter with BlueStyle and add it to tests

gdalle:
Other prominent JuliaCI repos like PkgTemplates.jl already use the [BlueStyle](https://github.com/invenia/BlueStyle). This PR does the same for BenchmarkTools.jl. More precisely, it
- sets up JuliaFormatter to use BlueStyle on this repo with the `.JuliaFormatter.toml` file
- applies automatic formatting to the existing code (mostly spacing modifications, all in the second commit)
- adds a test checking that formatting is correct in `test/runtests.jl`, along with a test-specific dependency in `Project.toml`
- adds a BlueStyle badge to `README.md`

### DIFF
--- a/.JuliaFormatter.toml
+++ b/.JuliaFormatter.toml
@@ -1,0 +1,1 @@
+style = "blue"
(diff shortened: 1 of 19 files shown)


gdalle:
Fails because of https://github.com/domluna/JuliaFormatter.jl/issues/746

PR #318, opened 2023-06-14
Don't tune manually set evals parameter

gdalle:
This PR add a boolean `evals_set` field to benchmark parameters for manually set evals. If `evals_set = true`, tuning does nothing.

Fixes #24 and related issues

PR #330, opened 2023-08-29
warm up at the beginning of every measurement

### LINKED ISSUE #291: Warmups are skipped when running a benchmark suite
If we tune the suite, then everything is warmed up all at once, but we don't get warmups before measuring individual benchmarks.
If we use cached parameters, then all the benchmarks are cold when we start the suite.

Ideally, we should warmup before running each individual benchmark in a suite.

willow-ahrens:
fixes #291


willow-ahrens:
@gdalle I think this one is ready for review!

gdalle:
Thanks @willow-ahrens! I'm a bit swamped at the moment but I haven't forgotten this, it's on my to-do list and then the 2.0 will probably follow

test/ExecutionTests.jl
Zentrik, line 1: After this `is_warm` I'd true right? If so does it make sense to test that.

Zentrik, line 1: Ditto

test/ExecutionTests.jl
```diff
@@ -148,12 +148,26 @@ tune!(b_pass)
...
+        sleep(2)
+        is_warm = true
+    end
+end
+
+w = @benchmarkable needs_warm()
+w.params.seconds = 1
+tune!(w)
+is_warm = false
+@test minimum(run(w).times) < 1e9
```
Zentrik, line 166: After this `is_warm` I'd true right? If so does it make sense to test that.

test/ExecutionTests.jl
```diff
@@ -148,12 +148,26 @@ tune!(b_pass)
...
+
+is_warm = false
 
-@test p.samples == 1
-@test p.evals == 1
-@test p.gctrial == false
-@test p.gcsample == false
+@test (@belapsed needs_warm() seconds = 1) < 1
```
Zentrik, line 170: Ditto

Zentrik:
Looks good to me

test/ExecutionTests.jl
willow-ahrens: `is_warm` would be true even before merging this PR (it becomes true if the function has been called at least once), so checking whether it is true would not test the behavior we want to add. I'd be happy to add a counter to the function that tests the number of executions, which should be at least 2, if you like?

Zentrik: Oh right sure, nevermind then.

Zentrik (approved):
LGTM

willow-ahrens:
@gdalle Hi there! Given this is on the release milestone and has an approving review, I'm sending a friendly ping your way. Thanks for all your work on this repo 😊

willow-ahrens:
Thanks for the prompt response! I believe not warming up is a correctness issue, especially since the package usually warms up but unexpectedly doesn't in certain circumstances. This is an important package in the Julia ecosystem so I'd like to see some fix to the issue. Perhaps we can ask someone else for help reviewing?

To answer your specific question:

`evals` is the number of times we run the function in each sample. https://github.com/JuliaCI/BenchmarkTools.jl/blob/42155f726e8da464baa5d917938a6d38fe20f7ae/src/execution.jl#L557C1-L559C20

When we call `@btime`, the function is first warmed up, then tuned, then run, so it would be executed at least 3 times. When we do `b = @benchmarkable ...; run(b)`, b is executed at least once. You're correct that I would be changing this behavior to execute b at least twice, once to warm up and once to measure. That's the behavior the documentation would lead one to believe is already in place, and best practice for benchmarking.

gdalle:
> You're correct that I would be changing this behavior to execute b at least twice, once to warm up and once to measure.

My issue is the case where the function destroys its input, so that only one evaluation is possible. Consider the following example:

```julia
julia> using BenchmarkTools

julia> @benchmark pop!(x) setup=(x=[0]) evals=1
 Range (min … max):  34.000 ns … 425.000 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     42.000 ns               ┊ GC (median):    0.00%
 Time  (mean ± σ):   42.534 ns ±   7.141 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

     ▃ ▄▅ ▆ ▇▇ █ ▇ ▅▄ ▃ ▁▁ ▁                  ▁ ▁              ▂
  ▄█▁█▁██▁█▁██▁█▁█▁██▁█▁██▁█▁▇▁▆▆▁▄▁▆▇▁█▁▇▁██▁█▁██▁█▁▇▁▇▇▁▇▁▇▆ █
  34 ns         Histogram: log(frequency) by time        69 ns <

 Memory estimate: 0 bytes, allocs estimate: 0.
```
Until my PR https://github.com/JuliaCI/BenchmarkTools.jl/pull/318 this used to error. With BenchmarkTools 1.4 it doesn't. With automatic warmup... it would again?

Zentrik:
You're example works with this pr, as we run a full sample to warmup (which includes the setup and teardown) not a single eval.

As for the part about this being a breaking change, would not removing `warmup` fix that. I don't see any pressing reason to remove it.

Thanks for all you work on this package.

willow-ahrens:
I understand your concern. However, I have tested this PR with your function and it does not error:
```
(@v1.9) pkg> dev /Users/willow/Projects/BenchmarkTools.jl/
   Resolving package versions...
  No Changes to `~/.julia/environments/v1.9/Project.toml`
  No Changes to `~/.julia/environments/v1.9/Manifest.toml`

julia> using BenchmarkTools

julia> @benchmark pop!(x) setup=(x=[0]) evals=1
 Range (min … max):   0.001 ns … 167.000 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):      0.001 ns               ┊ GC (median):    0.00%
 Time  (mean ± σ):   12.841 ns ±  19.438 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

  █                                                             
  █▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▄▆ ▂
  0.001 ns        Histogram: frequency by time           42 ns <

 Memory estimate: 0 bytes, allocs estimate: 0.
 ```

This is because my "warmup" step just re-uses the existing BenchmarkTools `samplefunc` that gets generated to benchmark the function, and `samplefunc` calls setup.

It's somewhat opaque because there is metaprogramming involved, but you can see `setup` pasted into the `samplefunc` definition here: https://github.com/JuliaCI/BenchmarkTools.jl/blob/c655edb69b652322ca41e70226b99069f297c5c2/src/execution.jl#L548

In BenchmarkTools.jl, we can understand the current `run(@benchmarkable ...)` behavior with the following pseudocode:

```
function samplefunc()
    setup()
    t = @elapsed begin
        for _ = 1:evals
            f()
        end
    end
    return t
end

for _ = 1:samples
    push!(measurements, samplefunc())
end
```

This PR is a petition to change the last three lines to:
```
samplefunc() #warmup
for _ = 1:samples
    push!(measurements, samplefunc())
end
```

**Comment by willow-ahrens** [comment]
@Zentrik I agree, we don't need to remove warmup. We may want to issue a deprecation though?

**Comment by gdalle** [comment]
Thanks for the clarifications, I had missed the setup phase in `samplefunc`. In that case, I think if we keep (but possibly deprecate) `warmup`, the PR is non-breaking and a strict improvement. If @Zentrik is confident in their ability to review the changes, I'll do a surface inspection myself and then probably merge

**Comment by gdalle** [comment]
> We may want to issue a deprecation though? Are deprecations breaking?

According to SciML ColPrac they are not

https://docs.sciml.ai/ColPrac/stable/#Incrementing-the-package-version

**Comment by willow-ahrens** [comment]
Okay, looks like this is all set then!

**Comment by Zentrik** [comment]
Cool, I'll try and test this out tomorrow.

File: src/execution.jl [hunk 3]
```diff
@@ -704,7 +708,6 @@ macro bprofile(args...)
     return esc(
         quote
             local $tmp = $BenchmarkTools.@benchmarkable $(args...)
-            $BenchmarkTools.warmup($tmp)
             $(hasevals(params) ? :() : :($BenchmarkTools.tune!($tmp)))
             $BenchmarkTools.Profile.clear()
             $BenchmarkTools.@profile $BenchmarkTools.run($tmp)
```
Zentrik, line 713: If `evals` is set, we don't run a warmup before profiling.

Zentrik (changes requested):

src/execution.jl
willow-ahrens: This PR adds the warmup to the `run` method, so it does get warmed up, but I see what you mean that the @profile now has no way to only measure the warm version.

willow-ahrens:
@zentrik I'm wondering if we should take a different approach here. Instead of modifying the behavior of `run` for `Benchmark`, could we modify the behavior of `run` for `BenchmarkGroup` to warm up before running each subbenchmark? The real problem here was that `run(BenchmarkGroup)` was not warming up each individual benchmark, so we could just fix the behavior of `run` on `BenchmarkGroup` to make a more minimal change.

Zentrik:
The problem I see with that approach is that someone might call `run(::Benchmark)`, e.g. you might have a `BenchmarkGroup` which you only run a single benchmark of. We probably want to run a warmup automatically in this case.

As an aside, do we intend to run the warmup twice. On master we call `warmup` once in the macro and then in `tune!`. And in this pr we warmup in `_lineartrial` and `_run`. I'm not sure I see the logic.

willow-ahrens:
We do intend to run warmup twice. We cannot guarantee that `tune!` is called just before `run`.

willow-ahrens:
I'll stick to this approach. Any ideas for how to fix the issue with profiling?

Zentrik:
I would assume something like `$(hasevals(params) ? :($tmp.samplefunc($tmp.quote_vals, $BenchmarkTools.Parameters($tmp.params; evals=1))) : :($BenchmarkTools.tune!($tmp)))` should work.

Zentrik:
> We do intend to run warmup twice. We cannot guarantee that `tune!` is called just before `run`.

Cool, I might make a pr to skip the second warmup then if we know tune! has been called.

willow-ahrens:
>      Cool, I might make a pr to skip the second warmup then if we know tune! has been called.
It may be difficult to know that it has been called.

Zentrik:
I figure for `@benchmark` we could just pass a boolean to skip the warmup or something when calling `run`. But I'll have a think.

Zentrik (approved):

willow-ahrens:
@gdalle okay, I think we're ready for your review, whenever you find time.

src/execution.jl
```diff
@@ -704,9 +708,17 @@ macro bprofile(args...)
...
-            $(hasevals(params) ? :() : :($BenchmarkTools.tune!($tmp)))
+            $(
+                if hasevals(params)
+                    :($tmp.samplefunc(
+                        $tmp.quote_vals, $BenchmarkTools.Parameters($tmp.params; evals=1)
+                    ))
+                else
+                    :($BenchmarkTools.tune!($tmp))
+                end
+            )
```
gdalle, line 719: Why is this only needed in `@bprofile`?

test/ExecutionTests.jl
```diff
@@ -148,12 +148,28 @@ tune!(b_pass)
...
+        sleep(0.1)
+    else
+        sleep(2)
+        is_warm = true
+    end
+end
+
+w = @benchmarkable needs_warm()
+w.params.seconds = 1
+tune!(w)
```
gdalle, line 166: Is there a way to check that `tune!` warms up too?

test/ExecutionTests.jl
```diff
@@ -283,7 +299,6 @@ io = IOBuffer()
 Profile.print(IOContext(io, :displaysize => (24, 200)))
 str = String(take!(io))
 @test occursin(r"BenchmarkTools(\.jl)?(/|\\)src(/|\\)execution\.jl:\d+; _run", str)
-@test !occursin(r"BenchmarkTools(\.jl)?(/|\\)src(/|\\)execution\.jl:\d+; warmup", str)
```
gdalle, line 286: Why does this disappear?

src/execution.jl
willow-ahrens: This was a bit of a compromise. `run` used to skip warmups. This PR changed it so that `run` now warms up. Unfortunately, `@bprofile` profiles `run`, and is now profiling the warmup run (potentially profiling compilation costs). My quick fix was to warm up before calling `run`, so `@bprofile` profiles a redundant warmup.

A more robust fix would be to give `run` the option to skip the warm up. This was suggested by @zentrik. This is a bigger change, but would not require `@bprofile` to change semantics.

test/ExecutionTests.jl
willow-ahrens: Sure! I can check that.

test/ExecutionTests.jl
willow-ahrens: This is left over from deleting warmup entirely from the package. It used to check that `@bprofile` wasn't profiling the warmup (important to avoid profiling the compiler).

willow-ahrens:
> I figure for `@benchmark` we could just pass a boolean to skip the warmup or something when calling `run`. But I'll have a think.

@Zentrik I think that to maintain semantics with how `@bprofile` used to work, we may want to do that boolean in this PR.

Zentrik:
I'm not sure this pr currently changes the semantic of `@bprofile` as it doesn't seem to me it should matter if there's an extra run in the profile. 
If you think its problematic feel free to fix it. Happy to remove all redundant warmups in a separate pr or in this one if you wish.

gdalle:
Neither of those is breaking as long as they don't change the semantics of existing code that does not use them

willow-ahrens:
In that case, @zentrik if you want to give it a go adding a boolean to the run kwargs for warmup, we could use that here to maintain semantics in bprofile.

willow-ahrens:
Okay, I added a `warmup` flag to `run` that allows us to skip warmups in certain limited scenarios, meaning that this PR preserves as much of the old execution semantics as possible. I think this is ready for review.

gdalle:
gotta love the commit messages ;) ping me when this is ready!

willow-ahrens:
@gdalle this one's ready! it's passing tests, but the icon is fail because the changes only reached 90% coverage rather than 91%.

Zentrik:
Thanks, sorry I've just been a bit busy and hadn't got around to this.

src/execution.jl
```diff
@@ -436,9 +444,8 @@ macro benchmark(args...)
     return esc(
         quote
             local $tmp = $BenchmarkTools.@benchmarkable $(args...)
-            $BenchmarkTools.warmup($tmp)
             $(hasevals(params) ? :() : :($BenchmarkTools.tune!($tmp)))
-            $BenchmarkTools.run($tmp)
+            $BenchmarkTools.run($tmp; warmup=$(hasevals(params)))
```
gdalle, line 448: Here IIUC `hasevals` checks whether `evals` has been set. Your reasoning is that if we have applied `tune!` before, we don't need a warmup. But in the situation where `evals` is set from the start, `tune!` does nothing and so warmup never runs.

src/execution.jl
```diff
@@ -656,9 +663,10 @@ macro btime(args...)
     return esc(
         quote
             local $bench = $BenchmarkTools.@benchmarkable $(args...)
-            $BenchmarkTools.warmup($bench)
             $tune_phase
-            local $trial, $result = $BenchmarkTools.run_result($bench)
+            local $trial, $result = $BenchmarkTools.run_result(
+                $bench; warmup=$(hasevals(params))
```
gdalle, line 668: same remark on tuning

src/execution.jl
```diff
@@ -704,10 +712,18 @@ macro bprofile(args...)
     return esc(
         quote
             local $tmp = $BenchmarkTools.@benchmarkable $(args...)
-            $BenchmarkTools.warmup($tmp)
-            $(hasevals(params) ? :() : :($BenchmarkTools.tune!($tmp)))
+            $(
+                if hasevals(params)
```
gdalle, line 716: same remark on tuning

test/ExecutionTests.jl
```diff
@@ -148,12 +148,38 @@ tune!(b_pass)
...
+w.params.seconds = 1
+
+#test that all measurements from lineartrial used in tune! are warm
+is_warm = false
+@test maximum(BenchmarkTools.lineartrial(w, w.params)) < 1e9
+
+#test that run warms up the benchmark
+tune!(w)
+is_warm = false
+@test minimum(run(w).times) < 1e9
```
gdalle, line 174: why not `maximum`?

test/ExecutionTests.jl
```diff
@@ -148,12 +148,38 @@ tune!(b_pass)
...
+
+w = @benchmarkable needs_warm()
+w.params.seconds = 1
+
+#test that all measurements from lineartrial used in tune! are warm
+is_warm = false
+@test maximum(BenchmarkTools.lineartrial(w, w.params)) < 1e9
+
+#test that run warms up the benchmark
+tune!(w)
```
gdalle, line 172: why `tune!` here?

test/ExecutionTests.jl
```diff
@@ -148,12 +148,38 @@ tune!(b_pass)
...
+is_warm = false
+@test minimum(run(w).times) < 1e9
+
+#test that belapsed warms up the benchmark
```
gdalle, line 176: should there be tests for the other `@b...` macros, which were also modified?

gdalle (commented):
Thanks for the contribution! I think most of the remarks stem from my own lack of understanding, but once we fix that it looks good to go

src/execution.jl
willow-ahrens: Yes, this confused me too a bit, but the situation is as follows:
If evals is set at macro expansion time, tune is not called, so we need to warmup, set warmup = true
if evals is not set at macro expansion time, tune is indeed called, so we don't need to warmup, set warmup = false

Note that we are interpolating whether params has evals into the quoted expression before tune runs.

src/execution.jl
willow-ahrens: I also think my same response applies here. Also note that both cases are also running in a macro.

src/execution.jl
willow-ahrens: In this case, hasevals is true, so we aren't calling tune, but we still want to warm up the code by running it once. We call 
`run($tmp, $BenchmarkTools.Parameters($tmp.params; evals=1); warmup=false)` to run one evaluation of the code, and we set warmup=false so as not to run it twice.

willow-ahrens: In this case, hasevals is true so we don't need to tune, but we still need to warm up. We call run with evals=1 to run one execution of the code, and somewhat confusingly set warmup=false because we don't need to run it twice. The reason for all this fuss is we're trying to point the @profile macro at an already warmed up call to `run`

test/ExecutionTests.jl
willow-ahrens: The only reason I called `tune!` here is that we never previously called tune on the benchmark. We tested that `lineartrial` warms up, but we never called `tune!`, and I'm used to calling `tune!` first. Would you rather I set evals explicitly here?

test/ExecutionTests.jl
willow-ahrens: This goes back to the reason I filed the original issue. If you have a benchmark that takes longer than 5 seconds to compile, then not warming up and calling run on the benchmark leaves you with a measurements vector whose only entry is measuring compile time. Since my code was reducing the benchmarks with `minimum` this effect was visible. However, I agree that this would be much clearer if we set it to maximum.

test/ExecutionTests.jl
willow-ahrens: We can easily test `@benchmark`. It would be harder to test `@btime` or `@bprofile` because their output is harder to parse in the test suite.

willow-ahrens: and @belapsed calls @benchmark, not sure which ones you'd like me to test.

test/ExecutionTests.jl
willow-ahrens: When we test the output of `@belapsed`, we are using exactly the same test as here, since that also calls minimum. We're just trying to figure out if only the cold execution was measured, the failure case I said above.

willow-ahrens:
@gdalle I responded to your comments. I don't think anything needs to change, but I'm happy to change anything you would like, just let me know.

test/ExecutionTests.jl
gdalle: ok let's keep it that way

gdalle:
Sorry this has stalled, trying to get it merged and release this week

test/ExecutionTests.jl
gdalle: no that's fine

gdalle (approved):

gdalle:
Thank you so much for your work and patience, and sorry for the delay. This is a really important change and I reviewed it to the best of my ability.
Julia users of the future, if everything collapsed, this is where it started

Issue #201, opened 2021-03-19
Setup for BenchmarkGroups


gustaphe:
When working with random numbers or would be useful to have a way to guarantee that each benchmark in a group uses the same series of numbers. Unless I'm mistaken there is no way to do this now. 

If `BenchmarkGroup` had a property `setup::Function` which was evaluated for each `Benchmark`, one could use `setup=()->seed! (1234)` to make sure that the benchmark runs use the same sets of numbers.

Or `setup`could be an `Expr` so you could create a new rng.

gdalle:
Wanna try a PR @gustaphe ?

PR #335, opened 2023-09-21
[WIP] Add seed to BenchmarkGroup

gustaphe:
Partial solution for #201 . Since it needs to be serializable, I opted for just a seed number. `seed!` is not super well documented, so `Int` may not be the perfect type, but it's pretty handy. Negative seeds appear to be invalid, so I could use that as a stand-in for `nothing`, as `Union{Int, Nothing}` ruined serialization.

* [x] PR
* [ ] Complete tests
* [ ] Documentation

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -7,6 +7,7 @@ JSON = "682c06a0-de6a-54ab-a142-c8b1cf79cde6"
 Logging = "56ddb016-857b-54e1-b83d-db4d58db5568"
 Printf = "de0858da-6303-5e67-8744-51eddeeeb8d7"
 Profile = "9abbd945-dff8-562f-b5e8-e1ebf5ef1b79"
+Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 UUIDs = "cf7118a7-6976-5b1a-9a39-7adc72f591a4"
 
--- a/src/BenchmarkTools.jl
+++ b/src/BenchmarkTools.jl
@@ -8,6 +8,7 @@ using Statistics
 using UUIDs: uuid4
 using Printf
 using Profile
+using Random
 
 const BENCHMARKTOOLS_VERSION = v"1.0.0"
 
--- a/src/execution.jl
+++ b/src/execution.jl
@@ -145,6 +145,7 @@ function Base.run(group::BenchmarkGroup, args...; verbose::Bool=false, pad="", k
         gcscrub() # run GC before running group, even if individual benchmarks don't manually GC
         i = 1
         for id in keys(group)
+            Random.seed!(group.seed >= 0 ? group.seed : nothing)
             @logmsg(
                 ProgressLevel, "Benchmarking", progress = ndone / nleaves, _id = progressid
             )
@@ -245,6 +246,7 @@ function tune!(group::BenchmarkGroup; verbose::Bool=false, pad="", kwargs...)
         gcscrub() # run GC before running group, even if individual benchmarks don't manually GC
         i = 1
         for id in keys(group)
+            Random.seed!(group.seed >= 0 ? group.seed : nothing)
             @logmsg(ProgressLevel, "Tuning", progress = ndone / nleaves, _id = progressid)
             verbose && println(pad, "($(i)/$(length(group))) tuning ", repr(id), "...")
             took_seconds = @elapsed tune!(
--- a/src/groups.jl
+++ b/src/groups.jl
@@ -16,12 +16,14 @@ makekey(v::Any) = string(v)::String
 struct BenchmarkGroup
     tags::Vector{Any}
     data::Dict{Any,Any}
+    seed::Int
 end
 
-function BenchmarkGroup(tags::Vector, args::Pair...)
-    return BenchmarkGroup(tags, Dict{Any,Any}((makekey(k) => v for (k, v) in args)))
+function BenchmarkGroup(tags::Vector, args::Pair...; seed=-1)
+    return BenchmarkGroup(tags, Dict{Any,Any}((makekey(k) => v for (k, v) in args)), seed)
 end
-BenchmarkGroup(args::Pair...) = BenchmarkGroup([], args...)
+BenchmarkGroup(args::Pair...; seed=-1) = BenchmarkGroup([], args...; seed)
+BenchmarkGroup(tags, args; seed=-1) = BenchmarkGroup(tags, args, seed)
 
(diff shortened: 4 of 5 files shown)


gustaphe:
* [ ] Don't rely on 1.9 kwarg syntax

(I thought about this and forgot to fix it)

gustaphe:
There is a problem in that this only works if the benchmarks take the same number of samples. I don't know if this should be enforced or marked as a caveat: "The same seed guarantees the same sequence of random numbers, but the benchmarks may still pick an unequal number of random numbers".

In practice, consistency *increases* even if it's not guaranteed to be perfect.

gdalle:
Probably a dumb question, but why not give a seed to each benchmarkable instead?

gustaphe:
I guess one could. But the seed needs to be set before an entire benchmark run, not for each sample. And if it's not for a group you can just run `seed!` before running the benchmark.

An alternative is to add a hook to `run`, like `run(group; setup=(seed!(1234)))` - I don't have a strong opinion.

gdalle:
Given this recent Discourse thread, I think a seed for each benchmarkable would also make sense:

https://discourse.julialang.org/t/making-benchmark-outputs-statistically-meaningful-and-actionable/104256/

What do you think @gustaphe ?

gustaphe:
I didn't think that would work, but it took me like a second to convince myself it would. If nobody's made a PR like that by Saturday I will

gdalle:
Wait, isn't it already possible to set the seed in the setup code of each benchmarkable?

gustaphe:
> Wait, isn't it already possible to set the seed in the setup code of each benchmarkable?

No. If you set the seed in the `setup`, it gets reset for every sample, not just at the start of the benchmark run. Consider 
```julia
julia> b = @benchmarkable sleep(rand([0, 0.5])) setup=(Random.seed!(1234))
Benchmark(evals=1, seconds=5.0, samples=10000)

julia> r = run(b)
 Range (min … max):  1.531 μs … 48.249 μs  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     1.739 μs              ┊ GC (median):    0.00%
 Time  (mean ± σ):   1.836 μs ±  1.147 μs  ┊ GC (mean ± σ):  0.00% ± 0.00%

         ▃▆██▅▄▃▁
  ▂▂▂▃▃▅▇█████████▇▆▅▄▄▃▃▃▃▃▃▃▃▂▃▂▃▃▂▃▂▂▂▂▂▂▂▂▂▂▂▁▁▁▂▂▁▂▁▁▁▂ ▃
  1.53 μs        Histogram: frequency by time        2.57 μs <

 Memory estimate: 192 bytes, allocs estimate: 5.
```
vs
```
julia> b = @benchmarkable sleep(rand([0, 0.5])) setup=(Random.seed!(1235))
Benchmark(evals=1, seconds=5.0, samples=10000)

julia> r = run(b)
BenchmarkTools.Trial: 10 samples with 1 evaluation.
 Range (min … max):  501.700 ms … 502.117 ms  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     501.726 ms               ┊ GC (median):    0.00%
 Time  (mean ± σ):   501.802 ms ± 156.542 μs  ┊ GC (mean ± σ):  0.00% ± 0.00%

  █           ▃
  █▇▇▁▇▁▁▁▁▁▁▁█▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▇▁▁▁▁▁▁▇ ▁
  502 ms           Histogram: frequency by time          502 ms <

 Memory estimate: 192 bytes, allocs estimate: 5.
```

each of these runs only ever use a single result from `rand`. What you want is a series of random numbers, but in a repeatable sequence.

gustaphe:
I would guess that the most useful use of this parameter is supplying it as `run(::BenchmarkGroup; seed=something)`, but this really looks like the best implementation.

docs/src/manual.md
```diff
@@ -85,6 +85,7 @@ You can pass the following keyword arguments to `@benchmark`, `@benchmarkable`,
 - `gcsample`: If `true`, run `gc()` before each sample. Defaults to `BenchmarkTools.DEFAULT_PARAMETERS.gcsample = false`.
 - `time_tolerance`: The noise tolerance for the benchmark's time estimate, as a percentage. This is utilized after benchmark execution, when analyzing results. Defaults to `BenchmarkTools.DEFAULT_PARAMETERS.time_tolerance = 0.05`.
 - `memory_tolerance`: The noise tolerance for the benchmark's memory estimate, as a percentage. This is utilized after benchmark execution, when analyzing results. Defaults to `BenchmarkTools.DEFAULT_PARAMETERS.memory_tolerance = 0.01`.
+- `seed`: A seed number to which the global RNG is reset every benchmark run, if it is non-negative. This ensures that comparing two benchmarks gives actionable results, even if the running time depends on random numbers. Defaults to `BenchmarkTools.DEFAULT_PARAMETERS.seed = -1` (indicating no seed reset)
```
gdalle, line 88: I wonder if this can have surprising side effects since it changes the global rng. It's a good thing that `seed = -1` disables it by default, I'm just trying to imagine a way this can come back to bite us

gdalle (approved):

rfourquet:
I'm not clear on the details here, but instead of asking for a seed, why not copy the state of the global RNG at the beginning `rng_copy = copy(Random.default_rng())`, and then before each benchmark run, restore this RNG state (`copy!(Random.default_rng(), rng_copy)`. This is way faster than seeding the RNG, and could even be done by default. If the use then wants to try a specific seed, then `seed!(seed)` can be done manually at the beginning.

Seelengrab:
I'd go one step further and let the user provide the entire `RNG` object they want to use, defaulting to a state-resetting copy of the default RNG.

gdalle:
@gustaphe what do you think of this proposal?

gustaphe:
I've been waiting for a good time to read up on this/try it out but haven't found it. The obstacles I see are:
* Serialization - as is, benchmark objects are serialized, by some method I don't quite understand, so using an integer worked out-of-the-box but the other methods (possibly even the suggestion of using `nothing` as a sentinel) don't.
* Is there a good way to supply an RNG object *and have it reset every time*? Because it's no use if this is just an RNG that is set once and then evolves throughout the experiment.
 
Like I said, I don't know if these are reasonable objections, I just haven't had time to think carefully about them.

Seelengrab:
Serializing an e.g. `Xoshiro` should be perfectly possible, it just needs to serialize the internal state & then reconstruct the RNG object from that state. I'm kind of surprised this isn't already implemented in Base.

PR #336, opened 2023-09-25
Reexport `std` and `var` from `Statistics`

milesfrain:
Fixes: #146 (part 2)

The original request was to just add a note in the docs about reexports from `Statistics`: https://github.com/JuliaCI/BenchmarkTools.jl/issues/146#issuecomment-1732649667

But then I found we were missing some re-exports that the docs assume exist. So I added those back in as originally proposed in the above issue.

This PR also  removes unnecessary `Statistics` imports in the tests. FYI @maleadt, This reverts your change from #294 to restore my change from #149.

If we don't want to go with the reexport strategy, that's also fine with me. I just want things to be consistent.

### DIFF
--- a/docs/src/manual.md
+++ b/docs/src/manual.md
@@ -353,13 +353,20 @@ BenchmarkTools.Trial
   allocs: Int64 19
 ```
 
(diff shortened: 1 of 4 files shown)


gdalle (approved):
Fine by me, thanks!

Issue #311, opened 2023-06-03
Memory estimate?


jariji:
```
BenchmarkTools.Trial: 4103 samples with 1 evaluation.
 Range (min … max):  1.112 ms …  1.475 ms  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     1.210 ms              ┊ GC (median):    0.00%
 Time  (mean ± σ):   1.213 ms ± 41.653 μs  ┊ GC (mean ± σ):  0.00% ± 0.00%

                   ▂▄▅▆▆█▆▇▅▄▄▅▂▂    ▁▃▄▃▇▄▄▅▄▁               
  ▁▁▂▂▃▄▄▄▅▅▃▅▄▆▅▆▇████████████████▇████████████▆▆▄▄▃▃▃▂▂▁▁▁ ▅
  1.11 ms        Histogram: frequency by time        1.31 ms <

 Memory estimate: 10.14 KiB, allocs estimate: 147.
```

I'm wondering why are the memory and allocations a single "estimate"? If the amount of memory usage varies over the samples, then statistics and a histogram could give a more complete picture.

Issue #277, opened 2022-02-19
Benchmarks only track minimum number of allocations and memory usage


Arkoniak:
It seems, that `BenchmarkTools.Trial` uses only single value for storing number of allocations and memory usage during trial runs. And judging from the results it track only minimum allocations and memory. Somewhat exaggerated example:

```julia
using BenchmarkTools

function f()
    n = rand() < 0.5 ? 1 : 100000
    return rand(n)
end
```

and result
```
julia> @benchmark f() evals = 1
BenchmarkTools.Trial:
  memory estimate:  64 bytes
  allocs estimate:  1
  --------------
  minimum time:     55.000 ns (0.00% GC)
  median time:      86.357 μs (0.00% GC)
  mean time:        56.463 μs (3.11% GC)
  maximum time:     573.892 μs (45.10% GC)
  --------------
  samples:          10000
  evals/sample:     1
```
which doesn't feel right.

PR #340, opened 2023-10-22
Default to median time, when for allocations

PallHaraldsson:


PallHaraldsson:
Note, idea take from @chriselrod's PR:

https://github.com/JuliaLang/julia/pull/51812
```
# macros are too awkward to work with, so we use a function
# mean times are much better for benchmarking than minimum
# whenever you have a function that allocates
function bmean(f)
  b = @benchmark $f()
  BMean(b)
end
```

chriselrod:
In my opinion, when you're comparing codes with different amounts of allocations, `mean` is the most appropriate. GC produces a skewed distribution, and `median` misses those heavy tails.

PallHaraldsson:
Yes, I accidentally used median, not mean. I meant to copy your idea as is. And I'm editing this blind. I've never made a macro or changed (or one that calls another macro...).

I get:
```
@btime sin(1)
ERROR: LoadError: UndefVarError: `prunekwargs` not defined
```

At least if people want this change, with mean, it would be good to get it confirmed, and may help. Or knowing if not so I can stop looking into this.

I've preferred the mean in some cases (though usually the min), but now I'm thinking was I wrong, when would that be best?

chriselrod:
> Yes, I accidentally used median, not mean. I meant to copy your idea as is. And I'm editing this blind. I've never made a macro or changed (or one that calls another macro...).
> 
> At least if people want this change, with mean, it would be good to get it confirmed. Or knowing if not so I can stop looking into this.
> 
> I've preferred the _median_ in some cases (though usually the min), but now I'm thinking was I wrong, when would that be best? Is it only shown in `@benchmark` to notice it's different from the mean?

I've encountered resistance to the suggestion before. Many argue minimum is the best, because all noise is positive, slowing you down from the ideal.
However, whenever heap allocations are involved, that's not really correct.

This is especially the case when you have a GC, but it is even true for `malloc`/`free`, but to a much smaller degree (`free` is normally extremely fast, but every now and then can trigger some work to make the freed memory actually available outside of fast reuse caches).

Ignoring this cost, which sporadically results in extra time required, is wrong.

chriselrod:
A short summary:
1. `minimum` assumes all noise is positive. Wrong when you incur costs that periodically show up as time, as then you can have negative noise as in "paid less of that cost than average".
2. `median` minimizes absolute error. The centering parameter of a Laplace/double exponential distribution is the sample median (the logpdf of this distribution sums the absolute value of errors).
3. `mean` minimizes square error. The centering parameter of a Gaussian distribution is the sample mean (the logpdf of this distribution sums the squares; this is often intuitive as it minimizes distance, while the central limit theorem explains why this one tends to work so well in practice).

What we care about generally is the expected value of the time.
If we run things a million times, the duration will be around a million times the expected value.

The [expected value](https://en.wikipedia.org/wiki/Expected_value) is generally best estimated by the sample mean, although distributional assumptions can change that (in which case the estimates disagreeing would probably be a reflection of that assumption being bad).
Some distributions don't have means, e.g. the Cauchy distribution is too variable and thus sample means have the same variability as individual samples. But I don't think we have that sort of extreme problem (but it can show up from things as benign as the ratio of two independent zero-mean gaussian variables).

gdalle:
Thank you for the contribution!
I'm a maintainer of BT but I really don't have the expertise necessary to debate on benchmarking methods.
However, as far as benchmarking interfaces are concerned, I don't really like the idea of outputting different things depending on what the benchmarked code does. I'll try to see if this has popped up in issues before.

gdalle:
See previous discussions:
- https://github.com/JuliaCI/BenchmarkTools.jl/issues/311
- https://github.com/JuliaCI/BenchmarkTools.jl/issues/277
- https://github.com/JuliaCI/BenchmarkTools.jl/issues/248

gdalle:
I just remembered where the other lengthy discussion about this was:
- https://github.com/JuliaCI/BenchmarkTools.jl/pull/258

gdalle:
Kristoffer was firmly against this and asking to name this new mean-focused macro differently.

KristofferC:
BenchmarkTools itself runs the GC explicitly (see `gcscrub`) so inferring things about the GC time consumption in a run is unlikely to be very useful. Also changing the statistics based on whether an allocation happens is weird.

If you want to get the mean, use `@benchmark` or introduce a  `@bmean` or something.

chriselrod:
I do support host adding `@bmean`.
No need to change how any existing stuff works.

> BenchmarkTools itself runs the GC explicitly (see gcscrub)

No idea how successful my solution is, but I do try to disable that:
https://github.com/chriselrod/ElrodConfig.jl/blob/a3f4303bea52489f1162d2a9a443c06fc90c4547/src/ElrodConfig.jl#L76

PallHaraldsson:
If we step back a bit, then the purpose of benchmarking, timing, isn't usually the timing as a number itself. It's can I improve it. Or have I reached the optimal code. Thus you may want to track it over time, but rather I would like a macro that compares two versions of my code and tells me which is faster.

>I do support host adding @bmean.
>No need to change how any existing stuff works.

I think that's actually worse. It may be helpful to Elrod or other users that think (or know) that the mean is better (those users can use `@benchmark`), but most would be confused: should I use `@btime` or `@bmean`?

I want the best info for `@btime`, condensed into one line, whatever that may be.

If you have to show only one number I think we all agree the minimum is best, at least when not allocating.

Do we think it would be good to show one more number, and possibly in same format as when not allocating?

```
@btime sin(1)
  1.877 ns (mean is 5% slower; 0 allocations)
```

The max is an order of magnitude slower always (ranges from 9x-14x), and at first I dismissed it as useless info (why even show it?), but maybe it isn't, showing speed from cold cache? Is that sometimes useful? At least it gets calculated into the mean. I'm not saying we should show it directly. The mean is seemingly mostly immune to the max as an outlier. The median even more, and I'm not sure we shouldn't be showing that rather than the mean.

Do you like two numbers as timings, or one as a fraction, as shown? Or do you insist one just one number?


[One other possible PR I'm not going to make would be showing e.g. 0–20 allocations, when there's a range. It's for rather unusual situations anyway, but the logic in this PR assumes 0 or not...]

PR #341, opened 2023-11-11
Update attribution in readme

LilithHafner:
Ref: https://github.com/oxinabox


gdalle:
I feel stupid pointing this out so please correct me if I'm wrong, but isn't this https://github.com/johnmyleswhite/Benchmarks.jl?

vchuravy:
@gdalle Yes you are correct.

LilithHafner:
Thanks, @gdalle! My bad.

PR #347, opened 2023-12-28
[WIP] Add experimental support for perf (via LinuxPerf.jl) 

Zentrik:


Zentrik:
Do we need to still support Julia 1.0 and 1.5?

Zentrik:
All tests should be passing now except for the type piracy tests. Do these methods need to be moved into `LinuxPerf`, how do I avoid type piracy?
https://github.com/JuliaCI/BenchmarkTools.jl/blob/b3f66b2608088115c028dda0355171e549da707b/src/serialization.jl#L35-L50

Zentrik:
Also right now it seems we run perf every sample but only store the last sample's result. Do we want to change this. If so do we want to store them all or should we only run perf on the last sample.

Zentrik:
I haven't added tests, but given the CI doesn't have perf available, not sure how useful it would be.

src/trials.jl
```diff
@@ -98,12 +136,16 @@ mutable struct TrialEstimate
...
 
 function TrialEstimate(trial::Trial, t, gct)
-    return TrialEstimate(params(trial), t, gct, memory(trial), allocs(trial))
+    return TrialEstimate(
+        params(trial), t, gct, memory(trial), allocs(trial), trial.linux_perf_stats
+    )
 end
 
+# Should we compare linux_perf_stats here?
 function Base.:(==)(a::TrialEstimate, b::TrialEstimate)
```
Zentrik, line 149: Should we check if `linux_perf_stats` are the same here? Given it's essentially a hidden field right now, I'd say you could argue we shouldn't.

DilumAluthge:
> I haven't added tests, but given the CI doesn't have perf available, not sure how useful it would be.

We might need to set up Buildkite CI on this repo. @vchuravy @staticfloat

src/trials.jl
```diff
@@ -8,9 +8,12 @@ mutable struct Trial
     gctimes::Vector{Float64}
     memory::Int
     allocs::Int
```
Zentrik, line 10: To limit how breaking this is, e.g. https://github.com/PTsolvers/HPCBenchmarks.jl/blob/a5985aaaf931efb0caf194d669e3bfcb90c5c08e/CUDA/common.jl#L24 would break, a constructor could be added which defaults `linux_perf_stats` to be nothing.

Zentrik:
I think main things left to do are:
- [ ] Passing custom options to Linux Perf doesn't work currently
- [ ] Add CI that can run with perf and tests

willow-ahrens:
I'm not sure I have the bandwidth as a maintainer of this package to support this kind of infrastructure directly in benchmarktools.jl. Is there a way that we can support this inside LinuxPerf.jl rather than in BenchmarkTools.jl? Maybe a separate `run_perf` function that runs the benchmarks in perf?

willow-ahrens:
it seems like too much overhead to me to support separate CI for this just to support this feature here, especially if it's possible to support this in linuxperf.jl. Could linuxperf support a separate `run_perf` function that runs a BenchmarkTools Benchmark through `perf`?

Zentrik:
I think if I can get this to a low enough maintenance overhead for you it would be best to add perf profiling in BenchmarkTools. I think we can just remove the extra CI and make it clear that this is experimental. I don't think there's much need to make sure the profiling works on CI and users can easily run their own tests. Let me know if there's anything else I can do to lower the overhead or if it's not going to be possible for the overhead to be low enough.

If you care why I don't think it makes sense to put this functionality in LinuxPerf or BaseBenchmarks, I've explained below.
So running perf on `samplefunc` seems to work ok, it has fairly high overhead but it seems to be not much noisier than this pr which is what really matters. However, actually integrating this in LinuxPerf or BaseBenchmarks seems difficult as presumably they would have to overload `BenchmarkTools._run` so that they can run perf whilst the function is still hot when benchmarking a BenchmarkGroup. I don't think we would want to do this by default for people using LinuxPerf so if I was to do this I would probably put it in BaseBenchmarks which doesn't have CI anyways. Also, they would have to setup a way to save and load the perf profiling results.

willow-ahrens:
I'm imagining something like:

```
run_perf(args...; kwargs..., posthook=f) = run_perf(args...; kwargs..., posthook=(b)->(run_perf_stuff(b); f(b)))
```

Zentrik:
This still needs tests and docs. I'd like to mark this as experimental or something so that we can make breaking changes to it without making a breaking change to BenchmarkTools.

I've moved the LinuxPerf stuff to https://github.com/Zentrik/BenchmarkToolsPlusLinuxPerf.

vchuravy:
Fairly late to the game here, but I would find something like this useful.
I do wonder if we need the code-duplication and if we maybe could use dispatch to configure the behaviour here.

willow-ahrens:
@vchuravy I agree, it would be nice to avoid code duplication, and I wonder if this PR could be achieved using functions as arguments.

Zentrik:
Is the code duplication concern referring to the changes in parameters.jl?
What do you mean by using functions as arguments?

willow-ahrens:
apologies for my delayed review as I finished up my graduate studies. I would also like to see linuxperf working here.

When I discuss duplicated code, I'm referring to how this PR uses several kwargs with similar names, such as 

```
run_customizable_func_only: Bool false
    enable_customizable_func: Symbol FALSE
    customizable_gcsample: Bool false
    setup_prehook: _nothing_func (function of type typeof(BenchmarkTools._nothing_func))
    teardown_posthook: _nothing_func (function of type typeof(BenchmarkTools._nothing_func))
    sample_result: _nothing_func (function of type typeof(BenchmarkTools._nothing_func))
    prehook: _nothing_func (function of type typeof(BenchmarkTools._nothing_func))
    posthook: _nothing_func (function of type typeof(BenchmarkTools._nothing_func))
```
as well as several repetitions of things like

```
if params.enable_customizable_func == :LAST
        params.customizable_gcsample && gcscrub()
        s = b.customizable_func(b.quote_vals, params)
        trial.customizable_result = s[1]

        if params.run_customizable_func_only
            return_val = s[end]
        end
    end
 ```
 
I don't know whether it is possible to make a simpler and hopefully more modular approach just yet, but I'd like to try!
 
By "using functions as arguments", I was insufficiently gesturing at something like the following approach (let me know whether you think this would work):

To benchmark the function `ex`, we start by constructing the samplefunc `f`:
```
function f(evals)
   res = nothing
   for i = 1:evals
      res = ex
   end
   return (result=res;)
end
```

Then we would wrap this in the standard timing infrastructure `g`:

```
function g(f)
   function inner(evals)
      tic = time()
      data = f(evals)
      toc = time()
      return (data..., time=toc-tic)
   end
end
```

Then, if a user wishes to produce a custom benchmarking function, they might do

```
function profile(f)
   function inner(evals)
      traces = @profile begin
         data = f(evals)
      end
      return (data..., profile=traces)
   end
end
```

And at the end of the day, we would have a BenchmarkResult with a namedtuple of attributes, which might include things like cpu counts and sampling profilers. Perhaps we could use a more involved approach using wrapper objects instead of wrapper functions, and call methods like `setup(::LinuxProfSampler)` or `sample(::LinuxProfSampler, evals)`, or possibly `finalize(::LinuxProfSampler, data)`, I'm not quite sure.

Would this kind of approach help the situation at all?

Zentrik:
No worries for the delay, thanks for the review, you do raise some good points. But anyways I'm fairly busy now so if we're going to move forward with #375 then I'm not going to bother working on this.

Thanks again for all the time you've put into this.

PR #350, opened 2023-12-31
Only run GC 3 times instead of 4 to scrub

Zentrik:
Per [slack](https://julialang.slack.com/archives/C03A7RP2ZMW/p1700484605291009?thread_ts=1700475300.698339&cid=C03A7RP2ZMW), 'Diogo changed it so it only needs 3 to promote everything to old generation now'.

### DIFF
--- a/src/execution.jl
+++ b/src/execution.jl
@@ -2,7 +2,9 @@
 # single sweep, since freeable objects may need more than one sweep to be appropriately
 # marked and freed.
 function gcscrub()
-    GC.gc(); GC.gc(); GC.gc()
+    GC.gc()
+    GC.gc()
+    GC.gc()
     @static if VERSION < v"1.10"
         GC.gc()
     end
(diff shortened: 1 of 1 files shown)


Diogo-Netto:
Olá boa tarde eu não entendi direito o sentido da notificação?

Em qua., 3 de jan. de 2024 14:08, Guillaume Dalle ***@***.***>
escreveu:

gdalle:
Sorry I tagged the wrong Diogo Netto 🤣

Diogo-Netto:
Sem problemas 😃

Em qua., 3 de jan. de 2024 14:50, Guillaume Dalle ***@***.***>
escreveu:

d-netto (commented):
If the goal of `gcscrub` is to ensure all objects reach the oldest generation, then this PR conforms with the changes we introduced in the GC in v1.10.

Want to note that since you are relying on internal GC behavior here (which we're free to change in any version), then it might be a good idea to guard these changes under a version check: i.e. put a `@static if VERSION=1.10` or alike.

gdalle:
@Zentrik wanna add that to make sure that on <= 1.9 we do 4 sweeps and on >= 1.10 we do 3?

gdalle (approved):

PR #351, opened 2024-01-04
Allow loading saved benchmark results from version <= 1.3.2

Zentrik:
After `evals_set` was added to the `Parameters` struct, it no longer was possible to load old files. Given the simplicity of the fix seems worthwhile.

I set the default to false, as that seems to replicate the behaviour of BenchmarkTools before this field was added.


gdalle (approved):
lgtm

PR #354, opened 2024-02-01
Deserialize Inf correctly in some cases

### LINKED ISSUE #353: cannot read benchmark JSON if run & written with `seconds=Inf`
```
using BenchmarkTools
suite = BenchmarkGroup()
suite["mwe"] = @benchmarkable sum(rand(10))
m1 = BenchmarkTools.run(suite; seconds=Inf)
BenchmarkTools.save("mwe.json", m1)
BenchmarkTools.load("mwe.json")
```

this is due to `convert(ft, fields[fn])` in `serialization.jl`

Zentrik:
Fix #353


adienes (approved):

gdalle (approved):
thanks!

PR #362, opened 2024-04-09
Support for named tuple destructuring

yha:
e.g.
```
@benchmark let 
    (; x, y) = $p
    f(x,y)
end
```
currently fails at macro expansion

### DIFF
--- a/src/execution.jl
+++ b/src/execution.jl
@@ -346,7 +346,12 @@ function collectvars(ex::Expr, vars::Vector{Symbol}=Symbol[])
         if isa(lhs, Symbol)
             push!(vars, lhs)
         elseif isa(lhs, Expr) && lhs.head == :tuple
-            append!(vars, lhs.args)
+            args = lhs.args
+            if !isempty(args) && isa(first(args), Expr)
+                # named tuple destructuring
+                args = first(args).args
+            end
+            append!(vars, args)
         end
     elseif (ex.head == :comprehension || ex.head == :generator)
         arg = ex.args[1]
--- a/test/ExecutionTests.jl
+++ b/test/ExecutionTests.jl
@@ -216,10 +216,12 @@ tune!(b)
     "good")
 @test_throws UndefVarError local_var
 @benchmark some_var = "whatever" teardown = (@test_throws UndefVarError some_var)
-@benchmark foo, bar = "good", "good" setup = (foo = "bad"; bar = "bad") teardown = (@test foo ==
-                                                                                          "good" &&
-    bar ==
-                                                                                          "good")
+@benchmark foo, bar = "good", "good" setup = (foo = "bad"; bar = "bad") teardown = @test(
+    foo == "good" && bar == "good"
+)
+@benchmark (; foo, bar) = (foo="good", bar="good") setup = (foo = "bad"; bar = "bad") teardown = @test(
+    foo == "good" && bar == "good"
+)
 
 # test variable assignment with `@benchmark(args...)` form
 @benchmark(


Zentrik:
Is it possible for after `args = first(args).args` for `args` to still be an `Expr`? Should we keep expanding till it's not?

I don't really know much about what this code is doing but it looks alright to me.

src/execution.jl
```diff
@@ -346,7 +346,12 @@ function collectvars(ex::Expr, vars::Vector{Symbol}=Symbol[])
         if isa(lhs, Symbol)
             push!(vars, lhs)
         elseif isa(lhs, Expr) && lhs.head == :tuple
-            append!(vars, lhs.args)
+            args = lhs.args
+            if !isempty(args) && isa(first(args), Expr)
```
Zentrik, line 350: Essentially, my point is should this be a while loop or something even more complicated or is it fine as is?

willow-ahrens:
This isn't even about the interpolation with dollar. For example, this also fails:

```
@benchmark let 
           (; x, y) = p
           f(x,y)
       end
```

yha:
> Is it possible for after `args = first(args).args` for `args` to still be an `Expr`? Should we keep expanding till it's not?
> 
> I don't really know much about what this code is doing but it looks alright to me.

I don't think that's possible, at least assuming valid Julia code is passed to `@benchmark`. I believe the `args` after that line can be an `Expr` only in a case like `(; x = (a,b), y)` which is not a valid lhs for assignment (and in that case the head of that would be `:kw`, not another `:tuple`, so it would require more complicated recursion if the goal is to make a better error message. But it's an unlikely mistake from the user anyway)

test/ExecutionTests.jl
```diff
@@ -216,10 +216,14 @@ tune!(b)
...
-    bar ==
-                                                                                          "good")
+@benchmark foo, bar = "good", "good" setup = (foo = "bad"; bar = "bad") teardown = @test(
+    foo == "good" && bar == "good"
+)
+if VERSION >= v"1.7"
+    @benchmark (; foo, bar) = (foo="good", bar="good") setup = (foo = "bad"; bar = "bad") teardown = @test(
+        foo == "good" && bar == "good"
+    )
+end
```
willow-ahrens, line 226: These tests are good, but if I understand correctly, I think this is missing the case where we interpolate a value from local scope with dollar, since that's part of the code that this PR changes.

src/execution.jl
willow-ahrens: Do we know if this is robust across julia versions? I'd like to see a more selective test, that perhaps checks the head of some of these exprs. to see that they are namedtuples.

e.g. probably we should see `:parameters` somewhere in here:

```
julia> dump(:((; x, y) = a))
Expr
  head: Symbol =
  args: Array{Any}((2,))
    1: Expr
      head: Symbol tuple
      args: Array{Any}((1,))
        1: Expr
          head: Symbol parameters
          args: Array{Any}((2,))
            1: Symbol x
            2: Symbol y
    2: Symbol a
    ```

willow-ahrens (changes requested):
This is a good start, just a few small tweaks! Thanks for your contribution!

test/ExecutionTests.jl
yha: I've added tests for interpolated values both for the tuple and named tuple case (though I don't think the code I changed is specific to interpolation in any way).

src/execution.jl
yha: Added more selective test, adding an error message in unrecognized cases.
The only cases I can think of that would initially parse but than trigger the error are the invalid forms `(; x, y=3) = z` and `(x, y; z) = q`.

willow-ahrens (approved):

willow-ahrens (changes requested):
Let's try to fix this for the general case

PR #365, opened 2024-04-26
Switch CI to buildkite

Zentrik:
For the LinuxPerf pr, to run tests with perf we need to use the Buildkite CI instead of github's.
I've split this change out from that pr.

For some reason pushing a new commit, doesn't cancel the old build. I think someone with access to the buildkite needs to fix that. Also, some parallelism would be nice.
The code coverage key needs to be provided as well.

Buildkite CI doesn't seem to support Julia 1.0, so we want to merge in #366 before this.


Zentrik:
Closed after discussion on slack that Buildkite doesn't have many runners and so this is going to slow down CI a fair bit. I've switched to only running buildkite CI on release julia and disabling the perf tests on Github actions in the linuxperf pr.

PR #366, opened 2024-04-26
Raise minimum Julia version to 1.6

Zentrik:
Split out from #347, LinuxPerf doesn't support 1.0 as it depends on PrettyTables which doesn't. Also, buildkite doesn't seem to support it either.
1.0 seems sufficiently old that it seems fine to me to drop support.

### DIFF
--- a/src/BenchmarkTools.jl
+++ b/src/BenchmarkTools.jl
@@ -12,7 +12,7 @@ using Profile
 const BENCHMARKTOOLS_VERSION = if VERSION >= v"1.9"
     pkgversion(BenchmarkTools)
 else
-    v"1.4.0"
+    v"1.6.0"
 end
 
 ##############
--- a/src/BenchmarkTools.jl
+++ b/src/BenchmarkTools.jl
@@ -8,12 +8,7 @@ using Statistics
 using UUIDs: uuid4
 using Printf
 using Profile
-
-const BENCHMARKTOOLS_VERSION = if VERSION >= v"1.9"
-    pkgversion(BenchmarkTools)
-else
-    v"1.6.0"
-end
+using Compat
 
 ##############
 # Parameters #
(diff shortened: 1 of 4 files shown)


willow-ahrens:
This seems like a good call, I don't think we have the investment to support anything earlier LTS (and if people want earlier versions they can use earlier versions of benchmarktools).

src/BenchmarkTools.jl
```diff
@@ -12,7 +12,7 @@ using Profile
 const BENCHMARKTOOLS_VERSION = if VERSION >= v"1.9"
     pkgversion(BenchmarkTools)
 else
-    v"1.4.0"
```
willow-ahrens, line 15: What is this line doing? is `pkgversion` supported on 1.6 and above? if so, perhaps we can simplify this to just call `pkgversion`.

willow-ahrens (commented):
looks good to me, apart from PkgVersion.

src/BenchmarkTools.jl
```diff
@@ -12,7 +12,7 @@ using Profile
...
+    v"1.6.0"
```
willow-ahrens, line 15: Let's replace these 5 lines with:

```
const BENCHMARKTOOLS_VERSION = VersionNumber(TOML.parsefile(joinpath(dirname(@__DIR__), "Project.toml"))["version"])
```

willow-ahrens (changes requested):
Let's take this opportunity to remove a hard-coded version string

src/BenchmarkTools.jl
Zentrik: Perhaps we should use Compat.jl instead? There seems to be some sort of locking going on in `pkgversion` (https://github.com/JuliaLang/Compat.jl/blob/220f4d7d34e9de9c14fece4b43425565713f2b25/src/Compat.jl#L442) and I don't understand it well enough to know if your change would be fine.

willow-ahrens: that would be fine as well! let's do that.

willow-ahrens (approved):

PR #367, opened 2024-05-02
Run gcscrub after warmup and remove unnecessary gc run for first sample

Zentrik:
Also, moved trial construction before gcscrub as it's mutable.

### DIFF
--- a/src/execution.jl
+++ b/src/execution.jl
@@ -109,14 +109,12 @@ end
 function _run(b::Benchmark, p::Parameters; verbose=false, pad="", warmup=true, kwargs...)
     params = Parameters(p; kwargs...)
     @assert params.seconds > 0.0 "time limit must be greater than 0.0"
-    warmup_start_time = Base.time()
     if warmup
         b.samplefunc(b.quote_vals, Parameters(params; evals=1)) #warmup sample
     end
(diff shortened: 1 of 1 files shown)


src/execution.jl
```diff
@@ -109,13 +109,14 @@ end
...
-    trial = Trial(params)
+    warmup_start_time = Base.time()
     if warmup
         b.samplefunc(b.quote_vals, Parameters(params; evals=1)) #warmup sample
     end
-    params.gcsample && gcscrub()
+    warmup_time = Base.time() - warmup_start_time
+    trial = Trial(params)
+    params.gctrial && gcscrub()
+    start_time = Base.time() - warmup_time
```
willow-ahrens, line 119: Based on https://github.com/JuliaCI/BenchmarkTools.jl/issues/368, I think we don't need to compute `warmup_time` and I think we can just do
`start_time = Base.time()`.

I hit this recently when I had a very long running first compile (~5s ugh) and then BenchmarkTools only measured one sample because it thought I was all set on benchmarking.

willow-ahrens (changes requested):
Pending the small change to what "counts" towards sampling time, this looks good to me.

willow-ahrens (approved):
looks good to me!

PR #373, opened 2024-09-03
no empty varargs

KlausC:
The call `max()` should not be defined in this module.
Reason is the sloppy use of `groups::BenchmarkGroup...` at a few spots in the `groups.jl`, specially when `first(groups)` is called later on.

This PR fixes that where applicable. It is no longer possible to call `min`, `max`, `ratio`, `judge` with an empty argument list.


Zentrik (approved):

willow-ahrens (approved):
This is great, thanks!

PR #380, opened 2024-11-18
"evaluations" -> "evaluations each" when printing `Trial`

asinghvi17:
This makes a bit more sense for first time users IMO, otherwise it's not obvious what a sample and evaluation are.


willow-ahrens:
I agree, this is a good fix to the wording.

willow-ahrens (approved):
Thanks for your contribution! I've changed the wording slightly to be even more specific.

PR #387, opened 2024-12-26
add documentation for evals_set parameter

### LINKED ISSUE #374: Unclear meaning of `samples` and `evals`
Hi there,

maybe I miss something completely, but I do not understand the meaning of `samples` and `evals` for the benchmarks.

If I run this code in REPL

```
using BenchmarkTools
BenchmarkTools.DEFAULT_PARAMETERS.samples = 1
BenchmarkTools.DEFAULT_PARAMETERS.evals = 1

foo_calls = 0

function foo()
  global foo_calls += 1
end

b = @benchmark foo()

@show foo_calls

b
```

it should, as I understood the documentation, evaluate `foo` once. However, it produces
```
foo_calls = 501494
BenchmarkTools.Trial: 1 sample with 993 evaluations.
 Single result which took 37.263 ns (0.00% GC) to evaluate,
 with a memory estimate of 16 bytes, over 1 allocations.
```

Why does it evaluate `foo` 993 times while it was called in reality 501494 times?

Priynsh:
Fixes #374  adding documentation for the evals_set parameter which seems to be missing otherwise.


willow-ahrens (commented):
This isn't an accurate description of this parameter: https://github.com/JuliaCI/BenchmarkTools.jl/commit/7d376340748edbb7eac06fc055dd32613869c019#diff-a4121e6af3eac1a9746b039d2ed91c9682bcfc0422ec37b531a0136d32498e39R82

If you're up for it, you can file a PR which removes the `evals_set` parameter, defaults evals to `nothing` in the Parameters constructor, and uses `tune` to update the value of evals only when it is set to `nothing`, then document the new sentinel value of `evals`.

Priynsh:
Thanks for the feedback! I have tried to implement the changes. Closing this PR and starting a new one.

PR #400, opened 2025-10-03
Support JSON 1.0 release

quinnj:


src/serialization.jl
```diff
@@ -24,7 +24,7 @@ function JSON.lower(x::Union{values(SUPPORTED_TYPES)...})
         field = getfield(x, i)
         ft = typeof(field)
         value = ft <: get(SUPPORTED_TYPES, nameof(ft), Union{}) ? JSON.lower(field) : field
-        d[name] = value
+        d[name] = value isa Float64 && !isfinite(value) ? nothing : value
```
quinnj, line 27: This was a surprise to me; I didn't realize JSON pre-1.0 allowed writing `Inf` out by default (which is against JSON spec) as `null`. I tried to think about other ways we could try and compat this, but I think this is simplest.

quinnj:
A fix has been merged upstream in JSON to avoid breaking this pre-1.0 compat issue

PR #419, opened 2026-04-10
Only GC when needed. Reduce allocs from mechanics. Add PrecompileTools workload.

IanButterworth:
Reduces harness overhead for benchmarks that don't allocate (the common case), and ships a precompile workload so the first `@benchmark` is fast.

### Changes

**Sampling hot path** (`src/execution.jl`)
- Write sample results into a `Ref{SampleResult}` instead of returning a 5-tuple — eliminates heap-allocating the return tuple on every sample (the samplefunc is stored as `Function`, so returns go through dynamic dispatch)
- Skip `gcscrub()` before each sample when the previous sample reported zero allocations — nothing to collect
- Skip `gcscrub()` before the trial when the warmup reported zero allocations
- `sizehint!` the `Trial` vectors after the first sample, estimating the number of remaining samples from elapsed time
- Avoid allocating a new `Parameters` for warmup; temporarily mutate `evals` and restore
- Use `resize!` instead of slice-copy (`estimates[1:completed]`) in `_lineartrial`
- Use explicit `push!(trial, s[1], s[2], s[3], s[4])` instead of `s[1:(end-1)]...` to avoid intermediate tuple allocation

**Result capture** (`src/execution.jl`)
- Add `capture_result` kwarg to `_run` — when `false` (the `run(::Benchmark)` path), the extra samplefunc call to capture the user's return value is skipped entirely
- `run_result` / `@btime` / `@btimed` still capture the return value via one additional call at the end

**PrecompileTools** (`src/BenchmarkTools.jl`, `Project.toml`)
- `@compile_workload` runs `@benchmark 1+1` and `show()` at precompile time
- Adds ~2s to precompilation but saves ~1.5s on every first use

**CI** (`.github/workflows/CI.yml`)
- Add `concurrency` / `cancel-in-progress` to avoid redundant CI runs

**Tests** (`test/ExecutionTests.jl`)
- Assert `gcscrub` is absent from the profile of a zero-allocation benchmark (GC skip works)
- Assert `gcscrub` is present when benchmarking `Ref(1)` (GC still fires when there are allocations)
- Assert the harness reports zero allocations for `sin(1)`

### Comparison vs main (Julia 1.12, Apple Silicon)

**Single benchmark** (`@benchmark foo($1)`)

|  | main | PR | speedup |
|--|------|----|---------|
| First run + display | 2.00 s (9.7M allocs, 474 MiB) | 0.50 s (1.5M allocs, 76 MiB) | **4×** |
| Second run (new expr) | 0.40 s (103k allocs, 3.6 MiB) | 0.03 s (1.3k allocs, 230 KiB) | **13×** |
| Steady-state | 0.40 s (102k allocs, 3.6 MiB) | 0.03 s (97 allocs, 172 KiB) | **15×** |

**BaseBenchmarks.jl** (scalar + tuple suites, `tune=false`)

|  | main | PR | speedup |
|--|------|----|---------|
| Wall time | 311 s | 94 s | **3.3×** |
| Allocations | 156M (6.0 GiB) | 41M (2.5 GiB) | **3.8×** fewer |
| GC time | 93.4% | 86.2% | |

The steady-state improvement comes from eliminating per-sample heap allocations (the `Ref` write-through) and skipping GC scrubs that were collecting nothing. The first-run improvement is from the PrecompileTools workload. The BaseBenchmarks speedup is dominated by the GC skip — most benchmarks in these suites don't allocate, so the old code was spending >90% of its time in unnecessary `gcscrub()` calls.

With help from Claude.


IanButterworth:
currently running "scalar" and "tuple" BaseBenchmarks takes 28s with this branch and 18s with main... So maybe the compilation overhead is too expensive

src/groups.jl
```diff
@@ -336,7 +336,7 @@ function Base.summary(io::IO, group::BenchmarkGroup)
     return print(io, "$(length(group))-element BenchmarkGroup($(tagrepr(group.tags)))")
 end
 
-function Base.show(io::IO, group::BenchmarkGroup)
+function Base.show(@nospecialize(io::IO), group::BenchmarkGroup)
```
IanButterworth, line 339: This is to make sure precompilation covers `IOContext{REPL.LimitIO{TTY}}`. Alternatively we'd make an extension on REPL just to precompile it... which seems excessive.

IanButterworth:
There's support here, I'm happy and a few different agents are too, so I'll go ahead and release a minor bump.

PR #431, opened 2026-07-30
Scrub conditionally and retry GC-contaminated samples

KristofferC:
`gctrial`/`gcsample` scrubs now run only when more than `SCRUB_BYTES` have been allocated since the last collection (default: `max(16 MiB, live heap ÷ 10)`). Any collection resets the counter, so a suite of small benchmarks shares one scrub instead of paying `gcscrub()` per benchmark; the group-level scrubs in `run`/`tune!` are conditional too. A scrub only pays off when real garbage has accumulated: below the threshold the next natural GC trigger is tens of MiB away and a full collection finds ~nothing, while costing 5–100+ ms.

A collection that still lands inside a sample and takes more than 10% of it gets one retry — the collection that fired is itself the scrub, so the retry starts with full allocation headroom. If the retry is also contaminated, the benchmark allocates past the headroom every sample: retries are disabled for the rest of the trial and GC remains part of its reported cost, as before. Below 10%, contamination is within run-to-run timing noise and the estimators absorb it.

Measured on BaseBenchmarks (`string`/`sort`/`collection`/`dates`/`misc`/`shootout`/`problem`, 625 benchmarks, `tune!` at `seconds=0.05` + `run` at `seconds=0.1`, Julia 1.12.6), together with the two-sweep `gcscrub` from #430:

|             | wall    | GC time       | full sweeps | maxrss   |
|-------------|---------|---------------|-------------|----------|
| main        | 401.6 s | 291.7 s (73%) | 2575        | 3118 MiB |
| this + #430 | 251.6 s | 138.3 s (55%) | 1179        | 3877 MiB |

This suite is allocation-heavy, so it is close to the worst case for the policy. The bigger win is for suites of cheap benchmarks that allocate only a little: a strictly zero-allocation benchmark already skips its trial scrub since #419, but a single allocation per evaluation re-enables it, so `main` pays a full `gcscrub()` per trial while accumulating almost no garbage — under this policy such benchmarks essentially never scrub. The threshold also closes the converse gap: `warmup_allocs` counts only the measured expression, so an allocating `setup` never triggered a scrub before, no matter how much garbage it accumulated. The maxrss increase is the expected tradeoff of fewer forced full collections; `SCRUB_BYTES` tunes it.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


IanButterworth:
SGTM. Not reviewed the code. @vtjnash fyi

